PR #28, opened 2026-03-26
feat!: update client, tools, and prompts for latest API spec

Kevin-layerV:
## Summary

- Updates the qurl-mcp server to match the current qurl-service API specification
- Adds 3 new MCP tools (`update_qurl`, `mint_link`, `batch_create_qurls`) for key agent workflows
- `update_qurl` is a superset of `extend_qurl` (extend expiration + tags + description + absolute expiry); `extend_qurl` is kept as a shorthand alias and its client method now delegates to `updateQURL` to avoid divergence
- Updates all types, prompts, and resources to reflect the new data model
- Aligns Zod schemas with the API's validation constraints (`max_sessions` ≤ 1000, `label`/`description` ≤ 500, tags regex + maxItems 10, empty-string guards on IDs)
- Fixes a real bug where `batchCreate` silently dropped per-item errors on HTTP 400 (API returns a `BatchCreateResponse` body on 400; the generic error path threw `HTTP 400` and lost the details)
- 240 tests passing (up from ~100)

### Breaking Changes

- QURL response shape changed: `qurl_link`, `metadata` removed from get/list (only `CreateQURLData` carries `qurl_link`); `tags`, `custom_domain` added
- Create endpoint path changed from `/v1/qurl` to `/v1/qurls`
- `description` replaced by `label` on create input (the API uses `label` on `CreateQurlRequest` but keeps `description` on `UpdateQurlRequest` — the MCP client mirrors that asymmetry)
- Quota response restructured with named `rate_limits` and `usage` objects; `plan` narrowed to `"free" | "growth" | "enterprise"`
- `QURL.status` narrowed to `"active" | "revoked"`; per-token `AccessToken.status` narrowed to `"active" | "consumed" | "expired" | "revoked"` (wider because tokens can expire or be consumed independently of the parent resource)

### Key API Changes Addressed

| Area | Old | New |
|------|-----|-----|
| Create endpoint | `POST /v1/qurl` | `POST /v1/qurls` |
| Response types | Single `QURL` type | `QURL` (get/list) + `CreateQURLData` (create, ephemeral `qurl_link`) |
| Create input | `description`, `metadata` | `label`, `access_policy`, `session_duration`, `custom_domain` |
| PATCH endpoint | `extend_by` only | `extend_by`/`expires_at` + `tags` + `description` (clear via `""` / `[]`) |
| Error format | `{ code, message }` | RFC 7807 Problem Details (with legacy fallback) |
| List filters | `limit`, `cursor` | + `status`, date filters, `sort`, `q` |
| Batch create | — | `POST /v1/qurls/batch` with per-item results (207 partial / 400 all-failed both pass through with structured per-item errors) |
| Mint link | — | `POST /v1/qurls/:id/mint_link` for generating additional access links on an existing resource |

## Test plan

- [x] `npm run build` passes
- [x] `npm run lint` passes
- [x] `npm test` — 240 tests, 16 test files, all passing
- [x] CI pipeline validates on PR


vikramlayerv:
## Code Review — 4-Agent Parallel Review

**Overall**: PR scope is well-executed — clean API migration, good test expansion (184 tests), solid Zod schema patterns, and proper breaking-change documentation. Findings below, grouped by severity.

---

### Bugs (3)

**1. `rotate-access.ts` — `label` doesn't exist in GET response**
Step 2 says "Note the `target_url`, `tags`, and `description`" but step 4 says "create a new QURL with the same `target_url` and `label`." `QurlData` (from `getQURL`) has no `label` field — that's a create-time input only. An agent following this literally will hallucinate a value or error. Also, `tags`/`description` captured in step 2 are never used in step 4.

**2. `update-qurl.ts` / `mint-link.ts` — uncaught `ZodError` from `.parse()` in handler**
The MCP SDK validates against `baseSchema` (no refinements), then the handler calls `updateQurlSchema.parse(raw)` which can throw a raw `ZodError` if mutual-exclusion fails. This propagates as an unhandled error instead of a clean tool error message. Should wrap in try/catch and return a user-friendly message.

**3. `client.ts` — `mintLink` sends `input ?? {}` defeating the `body !== undefined` fix**
The `request()` method was correctly fixed to use `body !== undefined`, but `mintLink` coerces `undefined` to `{}` before passing it, which means an empty JSON body is always sent even when the caller provides no input. Should pass `input` directly and let `request()` handle `undefined`.

---

### Concerns (8)

**4. `UpdateQURLInput` has `description` but create flow uses `label`** — Is `description` correct for PATCH, or should it be `label`? If the API truly uses different field names for create vs. update, add a comment. Otherwise this silently sends the wrong key.

**5. `audit-links.ts` — "Is it using a custom domain?" framed as an issue** — No explanation of why this is a problem. Agents will flag every custom-domain QURL as an issue. Should be "Verify custom domain is still valid" or removed from the flag list.

**6. `mint-link.ts` tool description doesn't mention `expires_in`/`expires_at` mutual exclusion** — `update_qurl` description says "Do not provide both" but `mint_link` says nothing. Asymmetric agent guidance.

**7. `secure-a-service.ts` promises security policy guidance but has no `access_policy` args** — The prompt's purpose is protecting a service, but `access_policy` (IP/geo/AI-agent controls) is never mentioned or offered as an arg.

**8. RFC 7807 error fallback chain is fragile** — `error.detail ?? error.message` skips `error.title`. If both `detail` and `message` are absent (valid per RFC 7807), the error message becomes the generic `HTTP ${status}`. Add `error.title` to the fallback chain.

**9. `list-qurls` handler — filter params not tested at tool level** — Schema accepts filters, but no handler test verifies they're forwarded to `client.listQURLs()`. Existing handler test only checks `limit` + `cursor`.

**10. `client.test.ts` — timestamp query param assertions are too weak** — `expect(calledUrl).toContain("created_after=")` passes even for empty values. Should assert with encoded timestamp values.

**11. `update-qurl.test.ts` — `issues[0]` access is fragile** — With multiple `.refine()` calls, Zod issue order can change. Use `.some(i => i.message === ...)` instead.

---

### Nits (4)

**12.** `QurlData`/`CreateQurlData` naming inconsistent with `QURL` convention used elsewhere (`IQURLClient`, `QURLAPIError`).
**13.** `sort` field in `list-qurls.ts` could be a `z.enum()` for better LLM discoverability.
**14.** `batchItemSchema` picks all 8 fields from `createQurlSchema` — effectively a no-op pick, brittle if schemas diverge.
**15.** Legacy error fallback test should assert `err.type === undefined` and `err.requestId === undefined`.

---

### Positives

- ✅ `body !== undefined` fix is correct and well-motivated
- ✅ RFC 7807 `type`/`instance` fields properly surfaced in `QURLAPIError`
- ✅ Base/refined Zod split for MCP registration is the right pattern
- ✅ `update_qurl` empty-update guard and mutual-exclusion refinements are clean
- ✅ Test coverage is thorough — schema edge cases, partial batch failures, error format migration all covered
- ✅ `mint_link` tip in rotate-access prompt is well-placed

justin-layerv:
## Review feedback addressed — commit 650734d

Thanks for the thorough reviews. Consolidated fixes below (triaged against `qurl/api/openapi.yaml` where applicable).

### Bugs fixed

1. **`rotate-access.ts` prompt referenced non-existent fields** — rewrote steps so only `target_url` is passed to `create_qurl`, then `update_qurl` restores `tags`/`description` on the new resource. Dropped `metadata` references (no such field on `QurlData` or `CreateQurlRequest`).
2. **Raw `ZodError` from refinement failures in `update_qurl`/`mint_link`** — handlers now `safeParse` the refined schema and return a clean `{ isError: true, content: [...] }` MCP result. Extracted into `zodErrorToToolResult` in `src/tools/_shared.ts` (the one real DRY opportunity in the diff).
3. **`mintLink` coerced `undefined` → `{}`** — defeated the `body !== undefined` fix in `request()`. Now passes `input` through directly; no body and no `Content-Type` header when the caller supplies nothing. Test updated.
4. **`QURL` interface had fields not in the API spec** — removed `metadata` (not on `QurlData`) and `qurl_link` (only on `CreateQurlData`). Narrowed `status` to `"active" | "revoked"`, made `custom_domain` nullable to match spec. `sampleQURL` and `get-qurl.test.ts` updated.
5. **RFC 7807 fallback skipped `error.title`** — chain is now `error.detail ?? error.title ?? error.message ?? HTTP ${status}`. Added a dedicated fallback test.

### Concerns addressed

6. **`audit-links.ts` framed custom domains as an issue** — reframed to neutral fact-gathering.
7. **`mint_link` description missing mutual-exclusion note** — added "Do not provide both expires_in and expires_at."
8. **`list_qurls` filter params untested at the tool layer** — added handler test verifying all filters forward to `client.listQURLs`.
9. **Timestamp query param assertions too weak** — replaced `toContain("created_after=")` with assertions on the fully URL-encoded values.
10. **`issues[0]` fragility in `update-qurl` schema tests** — switched to `.some(i => i.message === ...)` so refinement order changes don't break assertions.

### Nits fixed

11. **`CreateQurlData` → `CreateQURLData`** for naming consistency with `QURL` / `IQURLClient` / `QURLAPIError`.
12. **No-op `.pick()` in `batchItemSchema`** — dropped; batch items now use `createQurlSchema` directly so new create fields are picked up automatically.
13. **Legacy error fallback test didn't assert RFC 7807 fields are undefined** — added assertions for `type`/`instance`/`requestId`.

### Pulled in from "defer" list (per request)

- **`secure-a-service` now accepts access-policy args**: `ip_allowlist`, `ip_denylist`, `geo_allowlist`, `geo_denylist` (comma-separated strings, whitespace/empty-entries filtered), and `block_ai_agents` ("true"/"false"). The prompt builds an `access_policy` block inline for `create_qurl`. Five new tests cover the CSV parsing, geo/ip lists, and ai_agent_policy.
- **`batch_create_qurls` partial-failure UX**: the tool response now sets `isError: true` when `result.data.failed > 0`, and the serialized payload includes `request_id` from `result.meta` so agents can debug partial failures with the API provider. Tool description and tests updated.

### Deferred (with reasoning)

- **Concern about `UpdateQURLInput.description` vs `label`** — verified against `qurl/api/openapi.yaml`: `UpdateQurlRequest` uses `description`, `CreateQurlRequest` uses `label`. This is genuine API asymmetry, not an MCP bug. Current code matches the spec. No change.
- **`extend_qurl` still registered alongside `update_qurl`** — addressed in prior commit `c897764` with explicit shorthand documentation. Intentional surface, not dead code.
- **`listQURLs` truthy vs `!== undefined` guards** — empty strings have no semantic meaning for these query params; consistent with existing `cursor` handling. Skipped as non-bug.
- **`sort` as `z.enum()`** — API accepts flexible `field:direction` sort strings; enum would be restrictive. Current `z.string()` with a descriptive example is the pragmatic choice.
- **`AccessPolicy` type/schema drift risk** — architectural follow-up, not a defect. Out of scope for this PR.

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **215 passing** (up from 184), 16 test files
- `/simplify` pass: extracted the duplicated Zod→tool-result block into `_shared.ts`, deleted 4 WHAT-comments in `batch-create.ts` and `secure-a-service.ts`.

justin-layerv:
## Second round addressed — commit 3f3ca62

Picked up the follow-up review items including the nits.

### Fixed

1. **`secure-a-service` prompt was emitting `description` as a `create_qurl` param** — `createQurlSchema` has no `description` field, so Zod would silently strip it. The prompt now omits `description` from the `create_qurl` block and, when a description is supplied, adds a follow-up step instructing the agent to call `update_qurl` with the returned `resource_id` to set it. This matches the pattern already used by `rotate-access`.

2. **`mint_link` handler always sent an empty `{}` body when only `resource_id` was provided** — the destructured rest was `{}`, defeating the client's `body !== undefined` guard. The handler now detects the empty-body case and forwards `undefined`, so the client sends no body and omits `Content-Type`. Added a tool-level test for the `resource_id`-only path.

3. **`extend_qurl` example duration was `"168h"`** while every other tool uses `"7d"` — normalized for consistency. All duration examples across `create_qurl`, `update_qurl`, `extend_qurl`, and `mint_link` now use the same `"24h", "7d"` style.

### /simplify follow-up

The simplify pass flagged the first-pass `descriptionFollowUp` string as verbose — the parenthetical "(description is not accepted by create_qurl)" restated what the dev comment already says, and the LLM audience doesn't need the rationale. Tightened to a concise single-sentence instruction.

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **219 passing** (up from 215), 16 test files

justin-layerv:
## Seam audit — commit d0a0802

Cross-referenced every client method and schema in this PR against `qurl/api/openapi.yaml` and the live handler code in `qurl/internal/api/handlers`. Found one real bug plus several silent validation gaps.

### 🔴 Critical bug fixed

**`batchCreate` was dropping per-item errors on HTTP 400.** Per `qurl/internal/api/handlers/server.go:1126,1149`, when every batch item fails validation the API returns **400 with a full `BatchCreateResponse` body** (the per-item errors live in `data.results`), not an error envelope. The MCP client's generic `request()` error path saw no `json.error` field and threw a meaningless `QURLAPIError(400, "unknown", "HTTP 400")`, silently dropping all the structured per-item details that an agent would need.

Fix: `request()` now accepts a `passthroughStatuses: number[]` parameter. `batchCreate` opts HTTP 400 into pass-through, so the structured response flows through and the tool handler surfaces it with `isError: true`. Other 4xx/5xx codes still throw as before. Two new tests verify both the pass-through path and that non-400 batch errors still throw.

### 🟡 Schema validation tightened (was accepted locally, rejected by API)

| Field | API constraint | Was | Now |
|---|---|---|---|
| `create_qurl.max_sessions` / `mint_link.max_sessions` | 0–1000 | `.min(0)` | `.min(0).max(1000)` |
| `create_qurl.label` / `mint_link.label` | maxLength 500 | unvalidated | `.max(500)` |
| `update_qurl.description` | maxLength 500 | unvalidated | `.max(500)` |
| `update_qurl.tags` | maxItems 10, each 1-50, regex `^[a-zA-Z0-9][a-zA-Z0-9 _-]*$` | `z.array(z.string())` | full API constraints |

### 🔵 Type narrowing & doc consistency

- `QuotaOutput.plan` narrowed from `string` to `"free" | "growth" | "enterprise"` to match the spec enum.
- `get_qurl`, `update_qurl`, `extend_qurl`, and `mint_link` tool descriptions now document that the API accepts both resource IDs (`r_` prefix) and QURL display IDs (`q_` prefix) — per `openapi.yaml:2254` — and that the API resolves `q_` IDs to the parent resource automatically. Previously agents reading the descriptions wouldn't know they could use `q_` IDs. `DELETE /v1/qurls/{id}` keeps its narrower "r_ only" description since the spec explicitly restricts it.
- Shared resource_id description extracted to `describeResourceIdParam(verb)` in `tools/_shared.ts`.

### Verified correct (no action needed)

- `UpdateQURLInput.description` — confirmed present in `UpdateQurlRequest` at `openapi.yaml:2619`. The API is genuinely asymmetric: create uses `label`, update uses `description`.
- `CreateQURLInput` has no `description`/`metadata` — correct; the spec's create **example** (line 207-213) is stale but the actual `CreateQurlRequest` schema doesn't include them.
- `QURL.status` narrowed to `"active" \| "revoked"` — matches both the spec enum and the server code at `qurl.go:60-72` which explicitly states resources only have those two states (legacy "expired"/"consumed" are never set on new data).
- `qurl_link` removed from the `QURL` (get/list) interface — correct; only `CreateQurlData.qurl_link` exists.
- Bearer auth, `encodeURIComponent` on path params, RFC 7807 fallback chain including `error.title` — all clean.

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **232 passing** (up from 219)
- `/simplify` pass: extracted the repeated r_/q_ description blurb to `describeResourceIdParam()` in `_shared.ts` and tightened the batchCreate inline comment now that the `request()` JSDoc carries the full explanation.

justin-layerv:
## Third round addressed — commit 1b00c57

### 🔴 Bug fixed

**`rotate-access` prompt incorrectly claimed `label` was not accepted by `create_qurl`** — it is (per `createQurlSchema` in `create-qurl.ts:47`). My prior rewrite of this prompt lumped label in with tags/description, which would have misled agents into omitting a label on create and trying to set one via `update_qurl` (which doesn't accept label either). Corrected the parenthetical to only flag tags and description as the fields that need the follow-up `update_qurl` step.

### 🟡 Type tightening

**`AccessToken.access_policy`** — narrowed from `Record<string, unknown>` to the typed `AccessPolicy` interface so consumers get field-level type safety on nested policies.

**`AccessToken.status`** — narrowed from `string` to `"active" | "consumed" | "expired" | "revoked"` per `QurlSummary.status` in the OpenAPI spec. This enum is wider than the resource-level `QURL.status` (`"active" | "revoked"`) because individual tokens can be consumed or expire independently of the parent resource. Inline comment documents why the two enums differ.

### 🟡 Defensive validation

- New `resourceIdSchema(verb)` helper in `tools/_shared.ts` wraps `z.string().min(1).describe(...)`. Used by `get_qurl`, `update_qurl`, `extend_qurl`, `mint_link`. `delete_qurl` and `resolve_qurl` add `.min(1)` manually because their param shapes are distinct (r_-only and access_token respectively).
- `extend_by` on both `extend_qurl` and `update_qurl` now requires `.min(1)`.
- `access_token` on `resolve_qurl` now requires `.min(1)`.

Empty resource IDs would have hit `/v1/qurls/` with a missing path segment and produced a confusing API error; the client now rejects them at the schema boundary.

### 🔵 Consistency

**`client.mintLink` now uses `this.qurlPath(id)`** instead of inlining `encodeURIComponent`, matching the other methods and preventing future drift if `qurlPath` ever changes.

### /simplify pass

Three items caught and fixed:
1. **Dropped the `isError: false` additions on `mint_link`/`update_qurl` success paths** that I added in response to an earlier review. The reviewer's premise was factually wrong (`update_qurl` did not previously set `isError: false`), and adding it to just 2 of 9 tools creates inconsistency and implies significance where there is none. MCP defaults `isError` to false; all tools now consistently omit it on success and rely on the default.
2. **Inlined `describeResourceIdParam` into `resourceIdSchema`** since it had zero external callers after the helper was introduced.
3. **Tightened `update_qurl.extend_by` with `.min(1)`** for consistency with `extend-qurl.ts`, which the first round of this fix missed.

### Deferred (with reasoning)

- **`extendQURL` / `updateQURL` client method dedup** — already addressed in commit `c897764` with explicit "shorthand" documentation. Intentional surface.
- **`batchCreateSchema.url()` coupling** — reviewer explicitly labeled as "not an issue now, just noting".
- **`ListQURLsInput.status` narrowing** — intentional tradeoff to support CSV values per the API.
- **Duration string regex validation** (`/^\d+[smhd]$/`) — API is the ultimate validator and the describe strings already document the grammar; low priority per reviewer.

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **235 passing** (up from 232)

justin-layerv:
## Fourth round addressed — commit c8f1ac7

All three minor items from the latest review are in.

### Fixed

1. **`extendQURL` now delegates to `updateQURL`.** You were right to press on this — commit `c897764` only documented the tool-level relationship, not the client method dedup. Since `ExtendQURLInput` is a strict subset of `UpdateQURLInput` (just `extend_by`), TypeScript happily accepts the delegation with zero behavior change. The duplicated `request("PATCH", …)` call is gone, so if the PATCH endpoint ever needs a client-side tweak there's only one place to update.

2. **Trust-boundary comment on the error type cast** at `client.ts:273-275`. Acknowledges that the `json.error as { type?; title?; detail?; … }` cast is a trust-the-server cast, and notes that the fallback chain below tolerates missing fields — a malformed error response still degrades safely to `HTTP ${status}`.

3. **Empty-string rejection tests added to `extend-qurl.test.ts`** for both `resource_id` and `extend_by`, matching the parallel coverage in `get_qurl`, `update_qurl`, and `mint_link`. Catches a regression if `extend-qurl` ever diverges from the shared `resourceIdSchema`.

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **237 passing** (up from 235)

No deferrals this round.

justin-layerv:
## Fifth round addressed — commit 0eb3d51

All three minor hardening items from the latest review are in.

### Fixed

1. **Defense-in-depth shape guard in `batch_create_qurls` tool handler.** `batchCreate` opts HTTP 400 into the `passthroughStatuses` path, which contracts for a `BatchCreateResponse` body. If the API ever returns 400 with a different shape (e.g., a top-level malformed-request error), the handler would have crashed on `result.data.failed`. Now it flags unexpected shapes with `isError: true` and surfaces the raw response so agents get a real signal instead of silent mis-interpretation. New test covers this branch.

2. **Documented `ListQURLsInput` CSV / RFC-3339 conventions** in JSDoc on the TypeScript interface. The Zod `.describe()` strings already carry this information for MCP clients, but direct `QURLClient` consumers (bypassing MCP) only see the TypeScript type — the `status?: string` signature alone doesn't convey "comma-separated". Added JSDoc on `status`, the four date filters, `sort`, and `q`.

3. **Documented the distinct response shapes** on `IQURLClient.createQURL` and `getQURL`. `createQURL` returns `CreateQURLData` (carries the ephemeral `qurl_link`), `getQURL` returns the stable `QURL` shape (no `qurl_link`). The JSDoc now explicitly notes why they differ so consumers don't reach for a field that won't be there.

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **238 passing** (up from 237)

No deferrals. All reviewer suggestions (including those marked "low priority" and "just something to be aware of") are now addressed.

justin-layerv:
## Sixth round addressed — commit e6f83a8 + PR description update

### 🔴 Real bug fixed

**`updateQurlSchema` refinement rejected `description: ""`** — you caught a genuine edge case. The qurl API explicitly documents `description: ""` as the way to clear the field (`openapi.yaml` UpdateQurlRequest.description: "Pass an empty string to clear"), and `tags: []` as the way to clear all tags. The MCP refinement's "at least one update field required" check used a truthy `||` chain, which silently rejected empty strings. Switched to `!== undefined` checks so any explicitly-provided field counts, matching the API's clear semantics. The base-schema validators (`min(1)` on `extend_by`, `.datetime()` on `expires_at`) still reject empty values on those two fields, so the fix only opens up the intentional clear paths. Two new tests verify `{description: ""}` and `{tags: []}` both pass.

### 📝 PR description corrected

Updated the PR body to reflect reality:
- "Renames `extend_qurl` → `update_qurl`" → "Adds `update_qurl` as a superset of `extend_qurl`; `extend_qurl` is kept as a shorthand alias".
- Bumped the test count to 240.
- Rewrote the "Breaking Changes" section to reflect the actual state after all review rounds (status enum narrowing, plan enum, description/tags clear semantics, batch passthrough).

### ⏭️ Deferred (explicitly low priority / informational per reviewer)

1. **Duration format regex validation on `expires_in` / `extend_by` / `session_duration`** — reviewer explicitly labeled this a "conscious design choice" and noted that adding a regex "would risk rejecting formats the API actually accepts." Leaving the API as the ultimate validator.

2. **`isError` inconsistency (batch-create sets `false` on success; update/mint omit)** — reviewer explicitly labeled "very minor" and "functionally equivalent". I previously reverted `isError: false` on update/mint success paths based on /simplify feedback (adding it to only 2 of 9 tools created inconsistency with the rest). `batch-create`'s explicit `false` is meaningful because it's conditional on `failed > 0`, so leaving both as-is.

3. **`AccessPolicy` Zod schema + TS interface could drift** — reviewer explicitly labeled "reasonable trade-off" and informational. The current split gives cleaner types to direct `QURLClient` consumers who don't depend on Zod. No action.

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **240 passing** (up from 238)

justin-layerv:
## Seventh round addressed — commit 29f3f39

All four actionable items from the latest review are in.

### Fixed

1. **`linksResource` now documents the 50-item ceiling.** The resource description now reads "Snapshot of up to 50 active QURL links (most recent first). For full pagination use the list_qurls tool." Hoisted the limit into a named `LINKS_RESOURCE_LIMIT` constant so the description and the `client.listQURLs` call stay in sync.

2. **`list_qurls` `sort` param describe expanded.** The previous describe only showed `"created_at:desc"` as an example; now it enumerates the valid fields (`created_at`, `expires_at`) and directions (`asc`, `desc`, default `desc`) per the API spec. Agents no longer have to guess at the format.

3. **`secure-a-service` `one_time_use` coercion hardened.** Switched from the truthy check to `!== undefined`. Current behavior is unchanged (Zod enum guarantees non-empty string), but the refactored pattern is robust if the arg type ever changes from `z.enum(["true","false"])` to `z.boolean()` — an explicit `false` would have silently become `undefined` under the old pattern.

4. **`delete_qurl` `resource_id` now rejects `q_` prefixes at the schema boundary.** DELETE only accepts `r_` per the API spec; a `q_` ID would previously hit the API and get a confusing server-side rejection. Added a `.startsWith("r_", ...)` refine with a clear custom error message pointing callers to `update_qurl` or `mint_link` for q_ IDs. Tool description updated accordingly. Two new tests cover the `q_` rejection and empty-string rejection paths.

### Deferred (reviewer explicitly labeled each as "no action needed" / "fine as-is")

- **`extendQURL` client delegation observation** — reviewer: "fine as-is, just worth noting".
- **`listQURLs` `limit` asymmetry between MCP schema and direct client** — reviewer: "no action needed, just confirming the asymmetry is documented".
- **`batchCreate` → `createQurlSchema` coupling** — reviewer: "design call".

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **242 passing** (up from 240)

justin-layerv:
## Eighth round addressed — commit 1836c2c

### Fixed

**1. `.min(1)` on `expires_in` and `session_duration` in create_qurl and mint_link schemas.**

You caught a real bypass: `mintLinkSchema`'s truthy-based mutual-exclusion refine (`!(data.expires_in && data.expires_at)`) would silently pass `{expires_in: "", expires_at: "..."}` because `""` is falsy. The base-schema `.min(1)` now rejects the empty string before it reaches the refine. Applied the same treatment to `session_duration` in both files and to `expires_in` in `create_qurl` for defensive consistency — `update_qurl.extend_by` and `extend_qurl.extend_by` were already guarded this way.

**2. Regex format validation on `list_qurls.sort`.**

Previously deferred duration regex validation because the grammar wasn't documented, but `sort` is a closed set per the spec: `created_at|expires_at` field with an optional `:asc|:desc` direction. Catching malformed inputs at the MCP boundary is a free hardening win now that the describe text already enumerates the exact values. Pattern: `^(created_at|expires_at)(:(asc|desc))?$`.

### Tests (242 → 251)

- `list_qurls` sort validation: accepts field-only, accepts asc direction, rejects invalid field, rejects invalid direction, rejects wrong separator.
- `mint_link` rejects empty `expires_in` (covers the mutual-exclusion bypass path).
- `mint_link` rejects empty `session_duration`.
- `create_qurl` rejects empty `expires_in` and empty `session_duration`.

### Deferred (reviewer explicitly labeled)

- **`listQURLs` truthy query-param pattern** — reviewer: "probably the right behavior for query params (empty strings are meaningless)". Left intentional.
- **`delete_qurl` `startsWith`/`min(1)` ordering observation** — reviewer: "The ordering is correct. Just noting." No action.

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **251 passing** (up from 242)

justin-layerv:
## Ninth round addressed — commit 3eaf583

### Fixed

**1. `CLAUDE.md` architecture tree and tools table.** The tree was missing `src/tools/_shared.ts` and `src/tools/extend-qurl.ts`, and the tools table was missing `extend_qurl`. The tree now reflects all 10 files in `src/tools/` and the tools table now shows all 9 registered tools (matching what `server.test.ts` asserts). `extend_qurl` is explicitly marked as a shorthand alias for `update_qurl`. This fixes the orientation gap for future contributors and for Claude Code reading the file.

**2. `.min(1)` on `listQurlsSchema.status` and `q`.** Brought these in line with the defensive pattern applied everywhere else in the PR (resource IDs, durations, access tokens). The client's truthiness check in `listQURLs()` would have silently dropped an empty string anyway, but rejecting at the schema boundary gives agents a clearer error and eliminates the inconsistency.

### Deferred (both explicitly labeled by the reviewer as non-blocking)

- **`isError: false` on success paths** — reviewer: "Very minor." I've been through this loop a few times: `batch_create`'s conditional `false` is meaningful; `update_qurl` and `mint_link` omit it so all non-batch tools stay uniform with the rest of the codebase. A prior /simplify pass flagged adding explicit `false` as "implies significance where there is none". Holding.
- **`secure-a-service` description-without-target_url academic edge case** — reviewer: "purely academic" since `target_url` is required by the Zod schema.

### Validation

- `npm run build` — clean
- `npm run lint` — clean
- `npm test` — **253 passing** (up from 251)

Two new test cases cover the empty-string rejection on `status` and `q`.

PR #29, opened 2026-03-26
ci: add API spec drift detection workflow

Kevin-layerV:
## Summary

- Adds a scheduled GitHub Actions workflow that detects when the QURL API spec diverges from what the MCP tools are built against
- Stores the current API spec as a snapshot at `api-spec/qurls.yaml`
- Documents the maintenance process in CLAUDE.md

### How it works

1. Runs weekly (Monday 9am UTC) or on manual dispatch
2. Fetches the live OpenAPI spec from the configured URL
3. Diffs against the committed snapshot
4. If drift detected, opens a GitHub Issue with the diff (skips if one already exists)

### Configuration

- Spec URL is configurable via the `QURL_API_SPEC_URL` repository variable
- Defaults to `https://api.layerv.ai/v1/openapi.yaml`

## Test plan

- [x] Workflow YAML is valid
- [ ] CI pipeline validates on PR
- [ ] Manual `workflow_dispatch` trigger works after merge

### DIFF
--- a/.github/workflows/api-spec-check.yml
+++ b/.github/workflows/api-spec-check.yml
@@ -18,13 +18,14 @@ jobs:
   check-drift:
     name: Check for API spec drift
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Fetch current API spec
         id: fetch
         run: |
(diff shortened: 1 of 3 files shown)


justin-layerv:
## Review feedback addressed

All three optional observations fixed:

1. **Follow redirects** — added `-L` to curl so 3xx redirects are followed instead of silently skipped
2. **OpenAPI validation** — `head -1 | grep '^openapi'` → `grep -m1 '^openapi'` to handle specs with YAML comment preambles
3. **Update existing drift issue** — when an open `api-spec` issue already exists and the spec drifts again, a comment is posted with the updated diff instead of silently skipping

Also rebased onto current main.

PR #33, opened 2026-04-03
chore(client): simplify recent changes in qurl-mcp

justin-layerv:
## Summary

Simplification sweep across the 3 most complex files in the codebase. No behavior changes — all 149 tests pass before and after.

### Files changed

- **src/client.ts** — Extract `qurlPath(id)` helper for repeated URL pattern (3 call sites), extract `parseJSON` method from inline try/catch, use object spread for conditional Content-Type header
- **src/server.ts** — Replace 6 repetitive tool registration blocks and 2 resource registration blocks with loops; remove 6 now-unnecessary schema imports
- **src/prompts/secure-a-service.ts** — Replace 4 if-statement parameter checks with declarative filter/map over a parameter tuple array; build prompt text as single array join

## Test plan

- [x] `npm run build` passes
- [x] `npm run lint` passes
- [x] `npm test` — all 149 tests pass (13 test files)
- [x] No behavior changes — output is identical for all code paths

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -103,10 +103,8 @@ export class QURLClient implements IQURLClient {
     const url = `${this.baseURL}${path}`;
     const headers: Record<string, string> = {
       Authorization: `Bearer ${this.apiKey}`,
+      ...(body !== undefined && { "Content-Type": "application/json" }),
     };
-    if (body !== undefined) {
-      headers["Content-Type"] = "application/json";
-    }
 
     const response = await fetch(url, {
       method,
@@ -117,21 +115,11 @@ export class QURLClient implements IQURLClient {
     const text = await response.text();
 
     // Handle empty 2xx responses (e.g., 204 No Content from DELETE).
-    // Only deleteQURL hits this path — T is void there, so undefined is correct.
     if (!text && response.ok) {
       return undefined as T;
     }
 
-    let json: Record<string, unknown>;
-    try {
-      json = JSON.parse(text) as Record<string, unknown>;
-    } catch {
-      throw new QURLAPIError(
-        response.status,
-        "parse_error",
-        `Failed to parse response: ${text.slice(0, 200)}`,
-      );
-    }
+    const json = this.parseJSON(text, response.status);
 
     if (!response.ok) {
       const error = json.error as Record<string, string> | undefined;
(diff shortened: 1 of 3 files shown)


justin-layerv:
## Review feedback addressed

Both review comments have been fully addressed across two commits:

### 1. Restored explanatory comment on `undefined as T` cast (`97f0552`)

**Review feedback:** The removed comment on `src/client.ts` line 118 (`// Only deleteQURL hits this path — T is void there, so undefined is correct.`) was useful context for future readers about the non-obvious `undefined as T` cast.

**What changed:** Restored the comment at `src/client.ts:118`. The cast at the empty-response guard now has both lines of documentation explaining when and why `undefined` is correct.

### 2. Added `ToolFactory` type to guard tool registration loop (`b3f35a3`)

**Review feedback:** The loop-based registration relies on every tool factory returning `{ name, description, inputSchema, handler }` — an implicit contract that would fail at runtime rather than compile time if a factory returned a different shape.

**What changed:** Added a `ToolFactory` type alias in `src/server.ts:17-22` and applied it via `satisfies ToolFactory[]` on the factories array (line 38). This:
- Catches missing/misspelled properties at compile time (e.g., if a new tool factory omits `inputSchema`)
- Preserves inferred types in the loop body so `server.tool()` still gets specific handler signatures
- Uses `z.AnyZodObject` for the schema and `unknown` for the handler (handler type safety is provided by `server.tool()` at each call site, avoiding contravariance issues)

### Verification

- Build: clean
- Lint: clean
- Tests: 149/149 passing
- No security regressions (`encodeURIComponent` preserved, no hardcoded credentials, input validation unchanged)

justin-layerv:
## Review feedback addressed (round 3)

All three items from the latest review have been addressed:

### 1. Tightened `handler: unknown` to function type (`926e6ff`)

**Review:** `handler: unknown` in `ToolFactory` loses type safety — `server.tool()` would accept any value, including non-functions.

**Fix:** Changed `handler: unknown` → `handler: (...args: any[]) => Promise<{ content: Array<{ type: string; text: string }> }>` at `src/server.ts:23`. This validates the handler is a function returning the expected content shape. Args remain `any[]` because each tool has a different Zod-inferred input type (contravariance prevents using a concrete param type). Added eslint-disable comment and an explanatory note that exact handler signatures are validated by `server.tool()` at each call site via the `satisfies`-preserved inferred types.

### 2. Reverted Content-Type spread to explicit if block (`dab238d`)

**Review:** `...(body !== undefined && { "Content-Type": "application/json" })` relies on `...false` being a no-op — trips up readers and conflicts with the PR's simplification goal.

**Fix:** Reverted to the original `if (body !== undefined) { headers["Content-Type"] = ... }` block at `src/client.ts:107-109`. Straightforward over clever.

### 3. Added comment clarifying `one_time_use` ternary (`53d14f0`)

**Review:** The ternary-within-a-tuple for `one_time_use` is dense — `"false"` (truthy string) → `false` (boolean) is correct but non-obvious.

**Fix:** Added inline comment at `src/prompts/secure-a-service.ts:37`: `// "false" (truthy string) → false (boolean), undefined → omitted from output.`

### Verification

- Build: clean
- Lint: clean
- Tests: 149/149 passing

justin-layerv:
/simplify pass: code is clean, no changes needed.

Ran three parallel review agents (code reuse, code quality, code efficiency) across all changed files. 14 findings total — all triaged as false positives, style preferences, or suggestions that would revert intentional design decisions from prior review rounds:

- No duplicated utilities (qurlPath, parseJSON are novel)
- No unused imports or dead code (schemas still exported for tests)
- Comments flagged as "unnecessary" were specifically requested by reviewers
- Suggested handler type tightening already explored and rejected (contravariance)
- Prompt loop suggestion adds complexity (heterogeneous signatures) for 3 items
- Efficiency micro-optimizations (URLSearchParams allocation, array ops) are negligible for an API client

PR #52, opened 2026-04-17
ci: add npm dependency age check

justin-layerv:
## Summary
- Add npm dependency age check workflow (7-day quarantine on newly published packages)

Aligns with NHP repo security baseline. CodeQL already present.


justin-layerv:
## Review feedback addressed — all rounds consolidated

6 fixes applied across all reviews, nothing deferred. All fixes propagated to the 4 other npm dep-age PRs (qurl-typescript#32, qurl-portal#14, discord#21, website#189).

### Fixed

**1. Bug: `pipefail` + grep no-match crashes the workflow.** If lockfile changes don't include new `"resolved"` URLs (removing a dep, changing integrity hashes), `grep '"resolved"'` exits 1, `pipefail` propagates, `set -e` kills the script before the empty-check guard. Added `|| true` on the pipeline.

**2. `branches: [main]` added to pull_request trigger.** Every other PR-triggered workflow in this repo scopes to `branches: [main]`. Without it, this check fires on PRs targeting any branch.

**3. `$GITHUB_STEP_SUMMARY` table output.** Age-check results now surface directly in the PR Checks tab as a markdown table — package, version, age, and status per row. No more digging through logs.

**4. Package count logged upfront.** Prints "Checking N new/updated package(s)..." before the loop so reviewers know what to expect on large dependency updates.

**5. `--no-renames` on git diff.** Cheap safety net against rename detection confusion on large diffs.

**6. Checkout comment `# v6` → `# v6.0.2`.** Cosmetic consistency with all other workflows in the repo.

### Deferred

- **`workflow_dispatch` base_ref input** — current `main` fallback works fine for manual runs
- **Dep count threshold cap** — 5-min timeout with 0.2s rate limit supports ~1500 deps; premature
- **Abbreviated npm metadata endpoint** — doesn't include `.time`; no alternative available

justin-layerv:
## Review feedback addressed

2 fixes from the latest review. Applied to all 5 npm dep-age PRs.

### Fixed

**1. `curl -sSf` instead of `-sf` + `2>/dev/null`.** Transient failures (DNS, timeout, 403) now show errors on stderr as a breadcrumb in the Actions log instead of being silently swallowed. The `|| true` still prevents script abort.

**2. `--compressed` on curl.** Requests gzip/brotli from the npm registry for faster transfers on large package documents.

### Cross-repo

Same fixes applied to qurl-typescript#32, qurl-portal#14, discord#21, website#189.

justin-layerv:
Addressed all remaining CR feedback across the review rounds in `aca1a12` (including the previously-deferred nits).

### Fixed

**Substantive**
- ✅ **Silent-pass on full registry outage** (medium, from [4264802631](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264802631) #1): track `checked_count` vs `skipped_count`, fail when `checked=0 && skipped>0`. A green check with zero actual verifications is a false pass for a security control.
- ✅ **Registry response handling** ([4264495181](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264495181) / [4264802631](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264802631)): extracted `fetch_publish_time()` with explicit exit codes — 0 success, 1 = 404/410 (skip with warning: private/unpublished), 2 = transient failure (fail closed). Replaced the hand-rolled retry with curl's native `--retry 2 --retry-delay 2 --retry-all-errors` (more idiomatic, catches DNS + connection reset + 5xx per [4265053438](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4265053438) #5).
- ✅ **Package name validation** before `curl` interpolation. Defense in depth against crafted lockfile entries with path-traversal or query injection in the URL path.
- ✅ **`workflow_dispatch.inputs.base_ref`** ([4264474388](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264474388) #3, [4264774439](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264774439) #2, [4264802631](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264802631) #2): Previously deferred, now fixed. Manual runs from any branch get a useful diff instead of the silent no-op on `main`.
- ✅ **`jq` / `curl` availability sanity checks**: fail loudly if a future runner image change removes them rather than silently skipping checks.

**Cosmetic / documentation**
- ✅ Top-of-file comment framing the workflow's role relative to `dependency-review.yml` (known-CVE gate vs age gate) — [4264774439](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264774439) #5
- ✅ Inline comment noting the intentional unquoted `$new_packages` word-splitting — [4265053438](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4265053438) #2
- ✅ Inline comment flagging GNU-only `date -d` (runner is pinned to ubuntu-latest) — [4264774439](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264774439) #4
- ✅ Inline comment documenting the npm-alias / nested-package extraction limitation — [4264802631](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264802631) #3
- ✅ Inline comment justifying `fetch-depth: 0` (three-dot merge-base correctness) — addressed across reviewers
- ✅ Grouped `$GITHUB_STEP_SUMMARY` writes into heredocs / block redirects (shellcheck SC2129)

### Deferred

- **Full registry metadata fetch performance** ([4265053438](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4265053438) #1, [4264774439](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4264774439) #3): unavoidable — `.time` isn't on the abbreviated endpoint. The `--compressed` transfer + 5-min timeout + 0.2s sleep supports ~1500 packages, well above realistic lockfile size.
- **Dep count threshold cap**: premature. If a massive regeneration ever hits the timeout, we'll see it in logs before silent degradation, and the fail-on-all-skipped guard catches the interesting failure mode.

### Verification

- `actionlint .github/workflows/dependency-age-check.yml` clean
- `bash -n` passes on the extracted script (188 lines)
- Package-name validator tested against legit names (`express`, `@angular/core`, `lodash.get`, `legit-pkg.2` — all VALID) and injection attempts (`bad name`, `../etc/passwd`, `has spaces` — all REJECT)

justin-layerv:
Addressed the latest review round in `e4111a2` plus the parallel feedback from website#189 that applies here too.

### Fixed

- ✅ **Concurrency group** ([04:56 review](https://github.com/layervai/qurl-mcp/pull/52#discussion) #1) — scoped to `dep-age-${{ github.head_ref || github.ref }}` with `cancel-in-progress: true`. Rapid pushes to the same PR now cancel the superseded run instead of racing it through the sequential registry calls.
- ✅ **`trap ... RETURN` for temp file cleanup** ([04:56 review](https://github.com/layervai/qurl-mcp/pull/52#discussion) #2) — replaced per-branch `rm -f` in `fetch_publish_time()` with a single `trap`. Cleanup happens regardless of which return path the function takes, including unexpected exits between `mktemp` and the `case`.
- ✅ **Markdown injection hardening** (from the parallel website#189 feedback — same code pattern, same issue): added `safe_for_md()` and used it on `$pkg` in the invalid-name branch and on `$version` in all table-cell writes. Without this, a crafted lockfile with markdown syntax in a name or version would inject links/emphasis into the summary. Raw values still go to `::warning::` / `::error::` annotations (Actions logs, not markdown-rendered).

### Deferred

- **Large dependency update timing** (04:56 review #3) — informational, reviewer acknowledged the 5-min timeout + 0.2s sleep supports ~1500 packages. Premature to parallelize.
- **Nested dependency version resolution** (04:56 review #4) — already documented in the sed comment as a known limitation; reviewer also acknowledged npm v7+ hoisting makes it a non-issue in practice.

### Verification

- `actionlint .github/workflows/dependency-age-check.yml` clean
- `bash -n` passes on the extracted script
- `trap ... RETURN` is bash-only; runner is pinned to ubuntu-latest at job level (documented inline)

justin-layerv:
Addressed the latest [05:02 review](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4265611xxx) in `8faa489`.

### Fixed

- ✅ **`curl -sS ... 2>/dev/null` cleanup** (review #1): `-S` tells curl to show errors on stderr; `2>/dev/null` was throwing those away. Dropped the redirect so DNS failures, retry-exhausted transients, and unexpected HTTP statuses leave a breadcrumb in the Actions log. Added an inline comment explaining the flag combination so the next maintainer doesn't re-add the redirect "for consistency."

### Deferred (reviewer flagged as non-blocking)

- **Full package document fetch** (review #2): npm's `/{pkg}/{version}` endpoint returns the `package.json` for that version, which doesn't include `.time`. The full packument is the only public path to publish timestamps. Reviewer agreed this is "not needed now" — keeping the `--compressed` transfer as the optimization.
- **Explicit 429 handling** (review #3): `--retry-all-errors` already retries on 429 (and 5xx, DNS, connection reset). Combined with the inter-package `sleep 0.2`, this is adequate for realistic lockfile sizes. Would only matter at scale.

### 04:56 review

All four items (concurrency, `trap ... RETURN`, large-dep timing, nested-dep resolution) were already addressed in `e4111a2` — the latest review confirmed those fixes.

### Verification

- `actionlint` clean on `dependency-age-check.yml` (pre-existing SC2129 warnings in unrelated `api-spec-check.yml` unchanged).
- `bash -n` passes on the extracted script.

justin-layerv:
Addressed the [05:33 review](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4265726xxx) in `2ab566d`, plus brought this PR into parity with the latest hardening landed on the sibling dep-age PRs.

### Fixed (from the 05:33 review)

- ✅ **Heredoc → block-redirect form** (review #1): switched the step-summary header to `{ echo ...; echo ...; } >> "$GITHUB_STEP_SUMMARY"`. The heredoc actually rendered correctly (YAML dedent puts the body at column 0 before bash sees it), but the block-redirect form is self-documenting — won't confuse shellcheck/shfmt if anyone runs the extracted script outside YAML, and doesn't require tracing the YAML→bash dedent to know it works.
- ✅ **Alternative-registries comment** (review #2): extended the existing extraction comment with an explicit note that non-npmjs registries (GitHub Packages, private Verdaccio, etc.) are silently skipped by design.
- ✅ **Rate-limiting documentation** (review #3): added an explanatory comment at the `sleep 0.2` site — ~5 req/sec, 1500-package ceiling under the 5-min timeout, and `--retry-all-errors` handles any 429 that slips through.

### Fixed (parity with qurl-portal#14 and qurl-typescript#32)

- ✅ **`MIN_AGE_DAYS` input validator** — `[[ =~ ^[0-9]+$ ]]` guard right after the jq/curl sanity checks. A non-numeric `workflow_dispatch` input would have crashed later with a cryptic bash arithmetic error; now fails loudly with a clear message. Smoke-tested across the sibling PRs.
- ✅ **Expanded `SC2064` shellcheck-disable comment** — explains why early expansion is needed (`$response_file` is a `local` var that's out of scope by trap-fire time, so single quotes would break cleanup).

### Deferred

- **`trap ... RETURN` bash-specific warning** (review #4): already documented in the function's inline comment; reviewer explicitly noted "the comment is sufficient mitigation."

### Verification

- `actionlint .github/workflows/dependency-age-check.yml` clean.
- `bash -n` passes on the extracted script.

justin-layerv:
Addressed the [05:46 review](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4265765xxx) in `ed537a8`.

### Fixed

- ✅ **Scoped package URL encoding** (review #1): bash substitution `${pkg/\//%2F}` turns `@babel/core` into `@babel%2Fcore` before the curl call, matching the canonical registry API path. Both forms resolve today, but the encoded form removes the "works-but-undocumented" question. Unscoped names pass through unchanged.
- ✅ **Lockfile v3 assumption comment** (review #3): added a comment at the `jq .packages[...].version` lookup documenting that CI pins Node 20+ / npm 10+, so we always emit v3 lockfiles; a v1 lockfile would miss entries that only appear under `.dependencies`. Parity with qurl-portal#14 and qurl-typescript#32.
- ✅ **Quarantine window end date in error message** (review suggestion): error output for too-new packages now includes the eligible date:
  > `::error::${pkg}@${version} published 2 day(s) ago (minimum: 7; eligible after 2026-04-22)`
  
  Same info in the step-summary table row. Uses `date -u -d "@$eligible_epoch" +%Y-%m-%d` with a `|| echo "unknown"` fallback, consistent with the existing GNU-only `date -d` usage (already flagged at the top of the file). Developers now know when to retry without needing the bypass label.

### Deferred (reviewer marked as non-blocking)

- **Uppercase legacy package names** (review #2): reviewer confirmed the current strict regex is correct for the stated goal — npm has required lowercase for new packages since 2017, and any version bump to a legacy uppercase package surfaces as a skipped warning (not silently passing).
- **`sort -u` cross-version dedup** (review #4): reviewer called this a "nit just confirming correct behavior."
- **Exponential backoff with `Retry-After`** (suggestion): `--retry-all-errors` already retries 429s with a fixed delay. This repo's dep count makes sustained 429s unlikely.

### Verification

- `actionlint` clean on `dependency-age-check.yml`.
- `bash -n` passes on the extracted script.
- URL-encoding substitution tested against scoped (`@babel/core` → `@babel%2Fcore`) and unscoped (`typescript` → `typescript`) names.

justin-layerv:
Addressed the [05:56 review](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4265790xxx) in `4d3219f`.

### Fixed (from the 05:56 review)

- ✅ **Annotation injection via raw `$version`** (observation #1): sanitized `${version}` → `${safe_ver}` on every `::warning::` / `::error::` line that includes the version. `safe_ver` was already used for `$GITHUB_STEP_SUMMARY` cells, but the annotation lines were still consuming raw jq output. A crafted lockfile with a literal newline in a version string could have split the annotation line; low practical risk, trivial fix, matches the sanitization pattern used elsewhere.

### Fixed (parity with sibling PRs)

- ✅ **Real bug: silent-pass in the early-exit `git diff | grep -q .` check** (from discord#21 and qurl-typescript#32's latest rounds): same class of bug previously fixed on the main extraction pipeline. Under `pipefail`, a `git diff` failure gets inverted by `!` into the "No changes detected" branch, bypassing the security check. Split the diff from the filter with explicit error handling.
- ✅ **`workflow_dispatch.inputs.min_age_days`** (parity with qurl-portal#14 and qurl-typescript#32): added for per-run threshold override. `MIN_AGE_DAYS` now reads `inputs.min_age_days || '7'`.

### Confirmed non-issues (from the 05:56 review)

- **`curl` partial-success edge case** (observation #2): reviewer confirmed the `case 200)` match correctly rejects a split `http_code` like `"200\n000"` and falls through to the transient branch. Safe as-is.
- **Lockfile v3 assumption** (observation #3): reviewer confirmed the all-skipped guard correctly fails the run if a v1 lockfile ever appears.
- **`workflow_dispatch` + path filter** (observation #4): reviewer confirmed the early-exit handles this correctly.

### Verification

- `actionlint` clean.
- `bash -n` passes on the extracted script.
- Confirmed all six `::error::` / `::warning::` lines with a version now use `${safe_ver}` via grep.

justin-layerv:
Addressed via parity fixes in `1f30eaa`. The [06:06 review](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4265822xxx) was a pure "Approve" with no actionable items — but I pulled two parity items from the latest rounds on qurl-typescript#32 and qurl-portal#14 so this PR doesn't drift from its siblings.

### Fixed (parity with sibling PRs)

- ✅ **Dedicated `429)` case in `fetch_publish_time`**: emits a distinct `::warning::${pkg} — rate-limited by npm registry (429); --retry exhausted` before falling through to the same fail-closed transient path (return 2). `--retry-all-errors` already retries 429s; this case only fires after retries exhaust, helping diagnose rate-limit-driven failures vs generic network errors.
- ✅ **214-character length comment on `is_valid_pkg_name`**: added an inline note explaining 214 is npm's maximum package name length per the naming spec (includes the `@scope/` prefix).

### Confirmed non-issues (from the 06:06 review)

All five observations were reviewer-confirmed as correct/documented:
- **`trap RETURN` accumulation** (#1): "Purely theoretical for ubuntu-latest bash."
- **`sed` only covers `registry.npmjs.org`** (#2): documented and intentional (private registries can't be verified).
- **Lockfile v3 assumption** (#3): comment already correct for Node 20+/npm 10+.
- **`date -d` with untrusted registry data** (#4): "no command injection risk" — integer comparison after parse, fail-closed on weird dates.
- **Branch protection / required status check** (#5): out of scope for a code PR; that's a repo-settings decision to make alongside merge.

### Verification

- `actionlint` clean.
- `bash -n` passes on the extracted script.

justin-layerv:
Addressed the [06:15 review](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4265864xxx) in `eea52d4` plus picked up the "just past threshold" notice from qurl-portal#14 for sibling parity.

### Fixed (from the 06:15 review)

- ✅ **Raw `${version}` on the "OK" stdout log** (obs #1): swapped for `${safe_ver}` for consistency with every other output path. Reviewer noted injection risk is negligible (stdout not markdown-rendered), but the pattern is now uniform.
- ✅ **`MIN_AGE_DAYS=0` acceptance** (obs #2): tightened the validator from `^[0-9]+$` to `^[1-9][0-9]*$` so `0` is rejected explicitly. Zero-age quarantine is a no-op and would silently pass every package — worse than a misconfiguration error. Error message updated to "positive integer." Smoke-tested: `7`/`1`/`30` pass; `0`/empty/negatives/floats/text/injection-payloads all reject.

### Parity with qurl-portal#14

- ✅ **"Just past threshold" notice**: when a package passes but is `≤ MIN_AGE_DAYS + 1` days old (7-8 at default), emit a `::notice::` and mark the summary row with `:eyes: just past threshold`. Advance signal for deps that could flip to failing on a future PR. Three-branch age logic smoke-tested across 3/5/7/8/10/30 with `MIN_AGE_DAYS=7` — correct boundaries.

### Deferred (reviewer marked non-blocking)

- **Full package document fetch** (obs #3): reviewer explicitly said "not something to change" — `.time` is only on the full packument.

### Verification

- `actionlint` clean.
- `bash -n` passes on the extracted script.
- Security checklist in the review: every box still green.

justin-layerv:
Addressed the 06:29 review + brought this fully in parity with the latest sibling PRs (`qurl-portal#14`, `qurl-typescript#32`, `discord#21`, `website#189`).

**Fixed**
- 🐛 **429 warning swallowed by caller** — the review's explicit finding. `echo "::warning::..."` inside `fetch_publish_time()` was being captured by the caller's `$(fetch_publish_time ...)` substitution and never reaching the Actions log. Redirected to `>&2` so the annotation surfaces even on rate-limit exhaustion.
- `echo "$name_check" | grep -q .` → `[ -n "$name_check" ]` — fewer subprocesses, same semantics.
- `grep '^+' | grep '"resolved"'` → single regex `grep '^+.*"resolved"'`.
- `jq` version lookup now tolerant of a malformed lockfile via `|| true` + stderr suppression, so it yields a per-package skip instead of aborting under `set -e`.
- Added `shell: bash` to the step (explicit; `bash -n` still clean).
- Added `date -d @0` probe so BSD/macOS runners fail fast with a clear message rather than emitting cryptic date errors later.
- Documented the workflow + `age-check-bypass` label in `CLAUDE.md` (matching `qurl-portal`'s pattern).

**Deferred**
- Nothing — every item from the review and parity gaps identified against the sibling PRs has been addressed.

Validated with `bash -n` on the extracted step script; YAML parses cleanly.

justin-layerv:
Addressed all actionable nits from the [17:41 review](https://github.com/layervai/qurl-mcp/pull/52#issuecomment-4270170xxx) in `dc8a3ec`.

### Fixed

- ✅ **Nit #1 — `is_valid_pkg_name` regex**: expanded the allowlist to include uppercase and `~`. npm's `validate-npm-package-name` treats these as *warnings* (allowed for legacy packages), not hard errors — packages like `JSONStream` exist in the wild, and rejecting them here emits a misleading "does not match npm package name rules" for a legitimate entry. New regex: `^(@[a-zA-Z0-9~][a-zA-Z0-9~._-]*/)?[a-zA-Z0-9~][a-zA-Z0-9~._-]*$`. Smoke-tested — `JSONStream`, `@scope/Name`, `name~with~tilde` pass; `bad!name`, `bad(name)`, `path/traversal`, `@scope/name/extra` are still rejected. Shell metacharacters stay blocked.
- ✅ **Nit #2 — `trap` embedded single quotes**: `trap "rm -f '$response_file'" RETURN` → `trap "rm -f \"$response_file\"" RETURN`. Marginal (GNU mktemp doesn't produce single-quote paths) but defensive, matching the reviewer's exact suggestion.

### Parity (from sibling PRs, apply here too)

- ✅ **`workflow_dispatch` + `if:` behavior comment**: one-line explainer noting that manual runs always run (empty labels → `contains(...)` is false).
- ✅ **Expanded `|| true` comment on new_packages pipeline**: documents (a) grep's "exit 1 on no matches" contract and (b) `sort -u` dedup-by-name behavior for the same-package-two-versions edge case the reviewer raised as nit #3 (behavior is intentional).

### No-op (documentation already covers)

- **Nit #3 — same package, multiple new versions**: the reviewer was confirming behavior is correct; no code change. Added documentation inline so future readers see the rationale.
- **Nit #4 — lockfile v3 assumption**: already documented inline with the jq path. No change needed.

### Verification

- `actionlint` clean.
- `bash -n` passes on the extracted age-check script.
- YAML parses cleanly.
- Regex smoke-tested across 9 representative names.

justin-layerv:
No new review since 17:41 (fully addressed at 17:45 in `dc8a3ec`). Swept in two items from website#189's 17:56 round in `f018491`.

### Parity

- ✅ **Step-summary entries on early-exit paths**: both "no lockfile changes" and "no new packages (metadata-only diff)" branches now write a concise one-liner to `$GITHUB_STEP_SUMMARY`. Previously these paths left the Actions UI with an empty summary panel; now the outcome is visible without reading the full job log.
- ✅ **`trap ... RETURN` per-call lifecycle comment**: documents that each call installs a fresh trap which fires when that call returns, so reassigning the same trap name across calls is correct and doesn't leak. Behavior was already correct; sibling reviewer flagged it as a bash-ism worth calling out for future maintainers.

### Verification

- `actionlint` clean.
- `bash -n` passes on the extracted age-check script.
- YAML parses cleanly.

All prior items from the 17:41 review (regex uppercase + `~`, `trap` quoting, same-package/multi-version behavior doc, lockfile v3 note) remain addressed.

justin-layerv:
No new review since 17:41 (addressed at 17:45 / 18:08). Picked up two items from website#189's 18:13 round in `1605094`.

### Parity

- ✅ **Split `git diff` from the grep/sed/sort pipeline**: previously the single-pipeline form with a trailing `|| true` could silently swallow a diff or sed failure alongside the expected grep-no-match exit 1. Now `git diff` runs with its own error handler (`exit 1` on failure), and `|| true` on the pipeline scopes to grep/sed/sort only — where exit 1 genuinely means "no matching lines" (the metadata-only-diff case). Brings this into structural parity with the existing `name_check` split above it.
- ✅ **Actionable `date -d` error message**: improved from `"BSD/macOS not supported"` to explicitly cite the ubuntu-latest runner requirement + suggest `python -c '...'` as a portable replacement if a future caller needs macOS/BSD support.

### Verification

- `actionlint` clean.
- `bash -n` passes on the extracted age-check script.
- YAML parses cleanly.

All five sibling PRs now in sync at the latest canonical shape.

justin-layerv:
Parity with website#189 + qurl-typescript#32 + qurl-portal#14 18:36-round extraction in `300f8bc`. No new review since 17:41 on this PR.

### Fixed

- ✅ **Extracted dep-age check to `.github/scripts/check-dependency-age.sh`**: moves ~340 lines of inline bash out of the workflow's `run:` block. The script can now be linted with ShellCheck (verified clean) and tested in isolation with a local mock registry, rather than only via end-to-end CI runs. Workflow file shrinks from ~405 lines to ~71 (just trigger / permissions / env-block scaffolding). Behavior is byte-for-byte identical.
- ✅ **Documented "mixed public/private lockfile" edge case**: a lockfile with 1 public + 99 private packages will only age-check the 1 public one. The all-skipped-count guard catches the fully-private case but not this mix. Added as a known-limitation comment.

### Repo-specific preservation

- Kept the "CI pins Node 20+ / npm 10+" phrasing in the jq-path lockfile-v3 comment, matching this repo's Node-version pin (qurl-typescript says Node 22; qurl-portal says Node 18.17+/npm 9+ via package.json engines).
- Did **not** add the `git rev-parse --verify` base-ref pre-check that qurl-portal / qurl-typescript carry. Keeping the extraction faithful to this PR's prior inline body; happy to pick up as a follow-up if you want that improvement here too.

### Verification

- ShellCheck clean on `.github/scripts/check-dependency-age.sh`.
- `bash -n` clean.
- YAML parses cleanly.
- Script's own input validators (`MIN_AGE_DAYS` + `BASE_REF` required, positive-integer regex) give clear errors if invoked without the expected env.

justin-layerv:
No new review on this PR since 17:41 (addressed at 17:45 / 18:08 / 19:24 via the extraction). Picked up two parity items from the latest sibling sweeps in `601dad5`, including the `git rev-parse` gap I explicitly flagged as a follow-up in the `300f8bc` commit.

### Parity

- ✅ **`git rev-parse --verify` pre-check on the base ref** (previously deferred): produces a friendlier "Base ref not found" error than the bare `git diff` failure would, and closes a gap where the `|| true` on the later extraction pipeline could conflate a missing-base-ref failure with the expected grep-no-match exit 1. Parity with qurl-portal#14 / qurl-typescript#32 / website#189.
- ✅ **Enriched `2>/dev/null` comment**: matches website#189's fuller phrasing — the stderr suppression matches the per-package-skip posture elsewhere, so a failure here is expected if a future lockfile includes something jq can't parse.

### Verification

- ShellCheck clean.
- `bash -n` clean.

All five sibling scripts (website, qurl-typescript, qurl-portal, qurl-mcp, discord) are now at the same canonical shape. The only per-repo variation left is cosmetic (Node/npm version phrasing in the lockfile-v3 comment, which reflects each repo's actual pin).

justin-layerv:
No new review on this PR since 17:41. Picked up two parity items from the latest sibling sweeps in `4e83889`.

### Fixed

- ✅ **`codeql.yml` permissions scoped to the job** (parity with website#189 / qurl-portal#14): moved `actions: read` and `security-events: write` from workflow-level to the `analyze` job's own block, leaving only `contents: read` at workflow-level. No behavior change today (single-job workflow) but future-proofs: if a second job is ever added, it won't implicitly inherit CodeQL's SARIF-upload scopes.
- ✅ **`release-please.yml` SHA version comment**: `# v4` → `# v4.4.1`. SHA `5c625bfb5d1ff62eadeeb3772007f7f66fdcf071` is the v4.4.1 release commit (verified via `gh api repos/googleapis/release-please-action/git/refs/tags/v4.4.1`). Precise comments match the convention used elsewhere in this PR (`# v6.0.2`, `# v4.35.1`). Parity with qurl-typescript#32.

### Verification

- YAML parses cleanly on both files.
- ShellCheck clean on `.github/scripts/check-dependency-age.sh`.

No functional changes.

justin-layerv:
Superseded by the migration PR — all your discoveries are in ops-routines v0.1.2 with a 106-test harness. Closing without merging.

PR #71, opened 2026-04-27
feat!: add Dockerfile + Glama prep + defer QURL_API_KEY validation

justin-layerv:
## Summary

Adds a multi-stage Dockerfile + .dockerignore + glama.json + CI Docker build job so qurl-mcp can run as a container, claim ownership on Glama (`layervai/qurl-mcp` listing), and pass introspection probes from MCP directories. Also defers `QURL_API_KEY` validation from boot to first API call so the server can boot for introspection without a key.

## What this changes

**Dockerfile (multi-stage, hardened)**
- `node:22-alpine` build stage: `npm ci`, `tsc`, `npm prune --omit=dev`
- `node:22-alpine` runtime stage: ships `node_modules` (prod only) + `dist/`, copied with `--chown=node:node`
- Both stages pinned to a specific digest; Dependabot now tracks the docker ecosystem so the pin doesn't rot
- `USER node` (uid 1000) — non-root by default
- OCI labels: `org.opencontainers.image.source / description / licenses`
- `NODE_ENV=production`
- No baked secrets — `docker run -e QURL_API_KEY=lv_live_xxx ...` at runtime

**Deferred auth validation** (`src/index.ts`, `src/client.ts`)
- Server no longer hard-exits when `QURL_API_KEY` is unset. Instead it prints a stderr warning and boots cleanly so MCP introspection (`tools/list`, `resources/list`, `prompts/list`) succeeds. This is what Glama / Smithery / mcpservers.org probe for.
- `QURLClient.request()` throws `QURLAPIError(0, "missing_api_key", ...)` before any network call if the key is empty. Tool invocations surface the same clear, typed error rather than a server-side 401 that's harder to diagnose.
- Whitespace-only values (`QURL_API_KEY=" "`) take the same `missing_api_key` path via `.trim()` at config-read in both `index.ts` and the `QURLClient` constructor; constructor also null-guards undefined for dynamic callers.
- Both emit sites share `MISSING_API_KEY_MESSAGE` from `client.ts` — single grep target across logs, transcripts, and the CI probe.

**`glama.json`** — declares `justin-layerv` as the maintainer so Glama's "Login with GitHub to claim" flow can verify ownership of the org listing. Schema confirmed at https://glama.ai/mcp/schemas/server.json (single required field: `maintainers`).

**`.dockerignore`** — drops `.git`, `.github`, `.claude`, `.vscode`, `.idea`, tests, docs, env files, and the Dockerfile itself out of the build context. `package-lock.json` is kept (required by `npm ci`).

**CI: new `Docker` job** — builds the image with `docker/build-push-action`, then pipes a real MCP `initialize` + `tools/list` through `timeout 30s docker run -i` and asserts the JSON-RPC `name` field for every one of the 9 registered tools, plus that the introspection-only stderr warning prints. The `notify` Slack job depends on `docker` so a Docker build failure turns the message red.

**Test coverage** — `src/__tests__/client.test.ts` adds a `missing apiKey guard` describe block covering the empty-key, whitespace-only, and full-method-surface cases. Each asserts `QURLAPIError` with `code=missing_api_key`, `statusCode=0`, and that `fetch` was never called. 253/253 → **256/256**.

**Dependabot** — adds the `docker` package-ecosystem so the digest-pinned base image stays current.

## Verified locally

- `npm run build && npm run lint && npm test` — 256/256 pass
- `docker build .` — succeeds with the digest-pinned base
- `docker run -i --rm qurl-mcp` (no env): stderr warning prints, `initialize` succeeds, `tools/list` returns all 9 tools matched via the `"name":"<tool>"` strict grep
- `docker run --rm --entrypoint id qurl-mcp` → `uid=1000(node)`

## Test plan

- [x] Build, lint, test green locally
- [x] Docker image builds and passes MCP introspection without `QURL_API_KEY`
- [x] Container runs as non-root `node` user
- [x] New unit tests cover empty, whitespace, and per-method missing_api_key behavior
- [x] CI's Docker job runs the same probe Glama uses, asserting the full 9-tool surface via the JSON-RPC `name` field
- [x] Dependabot configured to track the Docker base image digest
- [x] CI green on this push
- [ ] Glama re-runs introspection and renders the score badge

BREAKING CHANGE: server no longer exits at boot when QURL_API_KEY is unset. Misconfiguration is now signaled via a stderr warning at startup and a typed `missing_api_key` error from every `QURLClient` request. Process supervisors that watched for non-zero exit on missing config will need to switch to detecting the per-call error or grepping the boot-time stderr warning. Some MCP hosts (notably older Claude Desktop releases) hide stderr — those users will see the failure only on first tool invocation, where the typed error message includes the actionable instruction to set the env var and restart.


justin-layerv:
Addressed both must-fix items from the review and most of the nice-to-haves in `cde334a`:

**Fixed**
- ✅ **Deferred auth validation** (review item #2). `QURL_API_KEY` check moved from boot-time hard exit to per-request throw in `QURLClient.request()`. Server now starts cleanly without the key, prints a stderr warning, and any tool invocation throws `missing_api_key` with a clear message. Removed the placeholder ENV from the Dockerfile.
- ✅ **Non-root user** (#1). `USER node` + `--chown=node:node` on the copied layers. Verified: `id` in container reports `uid=1000(node)`.
- ✅ **OCI labels** (#5). `org.opencontainers.image.source / description / licenses`.
- ✅ **NODE_ENV=production** (#7).
- ✅ **`.dockerignore` additions** (#10). Added `Dockerfile`, `.dockerignore`, `.vscode/`, `.idea/`.
- ✅ **CI docker build coverage** (follow-up #6). New `Docker` job in `ci.yml` that builds the image and pipes an MCP `initialize` + `tools/list` probe through `docker run -i` — same shape Glama uses. Catches alpine build drift in CI instead of at probe time.

**Verified locally**
- `npm run build && npm run lint && npm test` — 253/253 pass
- `docker build .` succeeds
- `docker run -i` with no `QURL_API_KEY` set: stderr warning, `initialize` succeeds, `tools/list` returns all 9 tools

**Verified `glama.json` schema** (follow-up review)
- `https://glama.ai/mcp/schemas/server.json` resolves and confirms `maintainers` is the documented field name. Single maintainer is fine for now; the layervai org currently has one human maintainer, so adding a duplicate would be cosmetic until the team grows.

**Deferred (non-blocking, reviewer flagged as such)**
- Base image digest pin — accepting `node:22-alpine` as the floating tag for now; can revisit when there's a Dependabot-driven cadence for it.
- `tini`/`--init` for PID 1 — for stdio transport the parent closes stdin, so this is academic. Punting until we have a real failure mode.

justin-layerv:
Addressed both blocking-ish items from the follow-up review in \`26ff6c9\`:

**(1) PR description updated** — now reflects what actually shipped (deferred auth, no placeholder ENV, stderr warning, per-call typed throw). Reviewers won't be chasing the removed approach.

**(2) Test coverage for the missing-key guard** — added a \`missing apiKey guard\` describe block in \`src/__tests__/client.test.ts\`:
- First test: \`new QURLClient({ apiKey: "", ... }).getQuota()\` rejects with \`name=QURLAPIError, code=missing_api_key, statusCode=0\`, and \`fetch\` is asserted to never be called.
- Second test: same guard exercised across all 10 public methods so the contract holds for the whole surface, not just \`getQuota\`.
- Tests: 253/253 → **255/255**.

**(3) CI now captures stderr and asserts the warning prints** — the Docker job redirects the container's stderr to \`/tmp/mcp-stderr\` and \`grep\`s for \`"QURL_API_KEY is not set"\` after the introspection assertion. A regression that strips the warning fails CI now.

**(4) Behavioral regression risk for hidden-stderr clients** — called out explicitly in the updated PR description under "Behavioral note for release notes" so it gets carried into the next release-please CHANGELOG.

**Nits**
- \`QURLConfigError\` subclass — agreed it'd be cleaner but flagged non-blocking; happy to follow up in its own PR if you want, otherwise leaving the documented \`statusCode 0\` convention as the seam.
- Cache mounts — optional speed-up; CI Docker build is already ~35s with no caching, so deferring.
- glama.json schema — confirmed resolves at \`https://glama.ai/mcp/schemas/server.json\` and \`maintainers\` is the only required field.

justin-layerv:
Addressed review items 1-4 in \`4ce9331\`:

**(1) Digest-pinned base image** — both Dockerfile stages now use \`node:22-alpine@sha256:8ea2348b068a9544dae7317b4f3aafcdc032df1647bb7d768a05a5cad1a7683f\`. Renovate/Dependabot will keep it current; meanwhile the supply chain seam is closed to match the SHA-pinned actions in the same workflow.

**(2) No-network assertion in the loop test** — added \`expect(mock).not.toHaveBeenCalled()\` after the loop. Now every public method is verified to throw *and* not issue a request, not just throw "some \`QURLAPIError\`."

**(3) Documented \`statusCode: 0\` sentinel** — added a JSDoc on \`QURLAPIError\` calling out that \`0\` means "client-side error before any network request." Future branching logic that inspects HTTP status will know to special-case it.

**(4) Unified the warning text** — extracted \`MISSING_API_KEY_MESSAGE\` constant in \`client.ts\`, used at both emit sites (\`index.ts:18\` boot warning and \`client.ts\` per-call throw). Single string for grep targets and the boot warning now inherits the actionable "Set it... and restart" phrasing the per-call message had. CI assertion still hits the substring \`"QURL_API_KEY is not set"\` so no workflow change was needed.

**(5) \`BREAKING CHANGE:\` footer** — agreed the boot-time \`process.exit(1)\` → \`stderr warning\` shift is a real behavior change for anyone with process supervision watching exit codes. Two ways to land it:

- **Squash-merge with a footer** — when squashing this PR, append \`BREAKING CHANGE: server no longer exits at boot when QURL_API_KEY is unset; misconfiguration is now signaled via stderr warning + per-call typed error.\` to the squash commit body. Release-please will pick it up and bump 0.2.x → 0.3.0.
- **Or amend the \`fix:\` commit to \`fix!:\`** before merge.

Either works and both are deliberate. Calling out here so it doesn't get lost at merge time. Will leave the choice to whoever lands the PR.

**Verified locally**
- 255/255 tests pass
- Docker rebuilt with the digest pin; introspection probe surfaces all 9 tools and the unified warning text

justin-layerv:
Addressed all three actionable items from the final review in \`cc9976d\`:

**(1) Probe asserts full tool surface** — CI now loops through all 9 registered tool names (\`create_qurl\`, \`resolve_qurl\`, \`list_qurls\`, \`get_qurl\`, \`delete_qurl\`, \`extend_qurl\`, \`update_qurl\`, \`mint_link\`, \`batch_create_qurls\`) and fails on any missing match. Catches the exact regression class (silently dropping a tool from registration) the reviewer flagged.

**(2) \`notify\` now depends on \`docker\`** — added \`docker\` to \`needs:\`, plumbed \`DOCKER_RESULT\` through to the Slack payload (overall status, per-job emoji, dedicated \"Docker:\" field). A Docker build failure now turns the Slack message red.

**(3) Whitespace-only \`QURL_API_KEY\` trimmed at config-read** — trim in both \`index.ts\` (boot-time) and the \`QURLClient\` constructor (defensive for direct callers). New unit test exercises \`apiKey: "   \\t\\n  "\` and asserts it takes the same \`missing_api_key\`/\`statusCode 0\`/no-fetch path as a truly empty key. 255/255 → 256/256.

**(4)** Re: \`MISSING_API_KEY_MESSAGE\` location — leaving it in \`client.ts\` for now; it's adjacent to the runtime guard and the boot-time emit imports it cleanly. If a third consumer shows up we can promote it to a shared \`messages.ts\` then.

**(5)** Re: \`BREAKING CHANGE:\` decision — leaning the same way the reviewer is. Pre-1.0 + exit-code-on-misconfig is a public contract that some users will have alerts on. Recommendation for the squash-merge body:

> \`feat!: add Dockerfile + Glama prep + deferred-auth refactor\`
>
> \`BREAKING CHANGE: server no longer exits at boot when QURL_API_KEY is unset. Misconfiguration is now signaled via stderr warning + per-call missing_api_key error from QURLClient. Process supervisors that watched for non-zero exit on missing config will need to switch to detecting the per-call error or grepping the boot-time stderr warning.\`

That bumps to 0.3.0 via release-please rather than 0.2.x, which feels right.

PR is otherwise ready from my end. CI on \`cc9976d\` is rerunning now.

justin-layerv:
Addressed both items from the latest review in \`315f296\`:

**(1) Dependabot now tracks docker** — added a third \`package-ecosystem: "docker"\` entry to \`.github/dependabot.yml\`. The digest-pin comment in the Dockerfile is now load-bearing on a real automation seam, not an aspirational one.

**(2) Null-guard \`config.apiKey\`** — \`(config.apiKey ?? "").trim()\` so a JS caller (or \`JSON.parse\`'d config) handing in \`undefined\` takes the typed \`missing_api_key\` path instead of a \`TypeError\` on \`.trim()\`. Comment updated to call out both reasons.

**(nit) PR description fixed** — the "lockfile-of-record" claim was wrong; \`package-lock.json\` is correctly kept (required by \`npm ci\`). Updated the description so the squash-merge body doesn't carry the false claim.

**Skipped on purpose**
- jq-based per-tool probe — current \`grep -q "\\"$tool\\""\` form is readable, the failure message is precise, and the only place a tool name appears quoted in \`tools/list\` is as a tool name. Cost > benefit.
- Tool-wrapper that catches \`code: "missing_api_key"\` and returns \`{ isError: true, ... }\` — agreed nice-to-have, but reviewer flagged it as out of scope and it's symmetric with how every other API error surfaces today. Filing as future polish.

256/256 tests pass; CI rerunning on \`315f296\`.

justin-layerv:
Addressed both must-fix items + two of the optional ones in \`d30a507\`:

**(1) Stricter tool grep** — switched from \`grep -q "\\"$tool\\""\` to \`grep -q "\\"name\\":\\"$tool\\""\`. Verified locally that all 9 tools match against the actual JSON-RPC \`tools/list\` response (the MCP SDK emits minified JSON so the literal \`"name":"<tool>"\` substring appears exactly once per registered tool). The \`update_qurl\`-mentions-\`extend_qurl\` regression class flagged in the review can no longer mask a dropped registration.

**(2) Stale test comment refreshed** — the comment now correctly describes the trim as the sole gate, not a non-existent "truthy check."

**(3) \`timeout 30s docker run\`** — entrypoint hangs (SDK notification-ordering bugs, etc.) now fail the step in 30s with \`Killed\` instead of running to the GitHub Actions default 6h. Cheap insurance.

**(4) Dropped the \`(PR #71)\` reference from the CI comment** — once merged the parenthetical is archaeological; the surrounding sentence already explains the deferred-auth context.

**Skipped on purpose (matches the reviewer's "optional" tag)**
- Multiple maintainers in \`glama.json\` — only one human admin on the org currently; will add when there's a real second person to list rather than padding it now.
- Dockerfile comment about \`package.json\` being copied for \`createRequire\` — already implicit from the existing \`COPY --chown=node:node --from=build /app/package.json ./\` line; reluctant to add comment-noise that explains what's already evident from the layout.

256/256 tests still pass; CI rerunning on \`d30a507\`.

**Operational follow-ups noted** (out of diff for this PR):
- Adding the \`Docker\` check to branch protection's required status checks — will handle alongside the merge.
- Squash-merge body should use \`feat!:\` + \`BREAKING CHANGE:\` footer for the boot-time exit-code shift, per release-please convention.

justin-layerv:
Resolved item #1 (the only blocker) plus #2 (cheap CI hardening) in the latest pushes:

**(1) Breaking change made unmissable at squash-merge time** — handled three ways so it can't slip:
- PR title renamed to **\`feat!: add Dockerfile + Glama prep + defer QURL_API_KEY validation\`** — the \`!\` survives the default squash-merge UI.
- PR body has a \`BREAKING CHANGE:\` footer at the bottom — release-please reads commit bodies for the trailer regardless of title prefix, so the bump to 0.3.0 is guaranteed via either path.
- Commit \`c090c73\` (this push) is unrelated, but I'll be flagging in the merge step that the squash-commit body should retain the footer.

**(2) \`set -euo pipefail\` + \`shopt -s inherit_errexit\` on the docker probe step (\`c090c73\`)** — addresses the bash command-substitution / \`set -e\` interaction. Now a \`timeout 124\` or \`killed 137\` from \`docker run\` aborts at the actual failure instead of leaking through to a misleading \"missing tool X\" message later. Ubuntu-latest runners have bash 5.x where \`inherit_errexit\` is standard.

**Skipped on purpose**
- **#3** \`statusCode: 0\` discriminator field — reviewer marked acceptable; agreed not worth restructuring \`QURLAPIError\` for one sentinel today. If a second pre-flight category lands, that's the right time to promote to \`kind: \"client\" | \"http\"\`.
- **#4** \`npm ci --ignore-scripts\` — reviewer flagged the babysitting cost ("skip if you don't want to babysit the flag against future dep changes"). Current deps don't use install scripts, but adding the flag means every future dep change has to re-validate that. Deferring until a real install-script attack vector lands.
- **#5a** \`normalizeApiKey\` helper — two call sites, both already do the same trim. Extracting a helper for two callers would invert the cost/benefit.
- **#5b** \`jq\`-based tool assertion — current grep relies on the SDK emitting minified JSON. Reasonable concern, but the SDK is locked at \`@modelcontextprotocol/sdk@^1.12.1\` and we'd see any pretty-print upgrade as a CI failure (which is exactly what we want — failure on SDK output change is detection, not silent pass). Considering this acceptable.

256/256 tests still pass; CI rerunning on \`c090c73\`. Should be the last round.

justin-layerv:
Applied all 5 follow-up items in \`4a58f54\`:

**(1) Tool-handler \`missing_api_key\` translation** — new \`withMissingApiKeyHandler\` wrapper in \`tools/_shared.ts\` catches \`QURLAPIError(0, "missing_api_key", ...)\` and returns \`{ isError: true, content: [...] }\` so MCP hosts surface the actionable error in the UI rather than as a transport-level JSON-RPC error (which some hosts hide). Applied to all 9 tool handlers. Other API errors (401, 404, 500, etc.) continue to throw — they carry statusCode/code metadata the host needs.

**(2) Per-method test tightened** — switched \`toBeInstanceOf(QURLAPIError)\` to \`toMatchObject({ code: "missing_api_key", statusCode: 0 })\`. Catches the regression class you flagged where a method routes around \`request()\` and throws a different \`QURLAPIError\` for unrelated reasons.

**(3) Whitespace coverage extended to all methods** — folded the standalone single-method whitespace test into the parameterized loop with \`["empty", ""]\` and \`["whitespace-only", "   \\t\\n  "]\` cases. Constructor's \`.trim()\` is now exercised against the entire 10-method client surface.

**(4) \`tini\` as PID 1** — added \`apk add --no-cache tini\` to the runtime stage and updated ENTRYPOINT to \`["/sbin/tini", "--", "node", "dist/index.js"]\`. \`docker stop\` now triggers a graceful shutdown via SIGTERM forwarding instead of relying on Node's PID-1 signal handling. Container still passes the full 9-tool introspection probe.

**(5) README Docker section** — added "Docker" section with build/run example, an explicit callout for the stderr warning + \`docker logs\` / \`2>&1\` recipes for hosts that hide stderr, and updated the Configuration table to reflect the deferred-auth model (\`Yes for tool calls\` rather than \`Yes\` blanket).

**New tests**
- \`src/__tests__/tools/_shared.test.ts\` — unit tests for the wrapper covering missing_api_key translation, non-missing-key QURLAPIError passthrough, non-QURLAPIError passthrough, and success passthrough.
- \`src/__tests__/tools/create-qurl.test.ts\` gains a smoke test asserting the wrapping is in place at the tool-handler level so a future refactor that drops the wrapper from one tool fails CI.

255/255 → **260/260** tests pass. Docker rebuild verified end-to-end:

\`\`\`
✓ create_qurl  ✓ resolve_qurl  ✓ list_qurls  ✓ get_qurl  ✓ delete_qurl
✓ extend_qurl  ✓ update_qurl   ✓ mint_link   ✓ batch_create_qurls
stderr: Warning: QURL_API_KEY is not set. Set it in the MCP server environment and restart to make API calls.
\`\`\`

CI rerunning on \`4a58f54\`.

justin-layerv:
Addressed the near-blocking item plus three of the polish items in \`0d783d3\`:

**(1) Resources now wrapped — UX is consistent across tools and resources.** New \`withMissingApiKeyResource\` helper in \`resources/_shared.ts\` mirrors the tool wrapper. Applied to both \`linksResource\` (\`qurl://links\`) and \`usageResource\` (\`qurl://usage\`). On a missing-key call the resource read returns:

\`\`\`json
{ "error": { "code": "missing_api_key", "message": "QURL_API_KEY is not set..." } }
\`\`\`

as JSON content, so MCP hosts render it through the same resource viewer they use for normal reads. No more divergence where \`tools/call\` got an \`isError\` content block but \`resources/read\` got a transport error.

**(3) \`QURL_API_URL\` trim** — symmetric with the apiKey path. A stray space in the URL now falls back to the default instead of producing a confusing fetch error.

**(4) CI probe now exercises \`tools/call\` under no-key mode** — extended the docker probe with a fourth JSON-RPC message:

\`\`\`json
{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"list_qurls","arguments":{}}}
\`\`\`

and asserts both \`\"isError\":true\` and the actionable \`QURL_API_KEY is not set\` substring in the response. Closes the loop on the user-facing contract — the wrapper-applied invariant is now exercised end-to-end through the same JSON-RPC surface a third-party probe would use, not just at the unit-test level.

**(6) README phrasing** — \"Conditional (see description)\" with the introspection-vs-tool-call distinction in the description column.

**Skipped (matches reviewer's non-blocking tag)**
- **(2) \`statusCode: 0\` discriminator** — JSDoc on \`QURLAPIError\` already documents the sentinel; reviewer agreed current shape is consistent with HTTP \"status 0 = no response\" convention. If a second pre-flight category lands, that's the right time to promote to a discriminator.
- **(5) Dockerfile \`apk add\` ordering** — reviewer marked \"not material\"; current ordering reads top-to-bottom and matches the \"add OS packages, then app code\" convention.

**Tests added**
- \`src/__tests__/resources/_shared.test.ts\` — unit tests for the resource wrapper (translation, non-missing-key passthrough, non-error passthrough, success passthrough).
- \`src/__tests__/resources/links.test.ts\` — smoke test asserting the wrapping is in place at the resource-handler level.

260/260 → **265/265** tests pass. Docker probe extended to 4 RPC steps; all new assertions verified locally.

justin-layerv:
Closed the coverage gap in \`196a91f\`:

**Coverage gap (LGTM-modulo)** — new \`src/__tests__/wrapper-coverage.test.ts\` parameterizes the wrapper assertion across **every tool handler (9) and every resource handler (2)**. Each case stubs the relevant client method to throw \`missing_api_key\` and asserts the handler returns the documented \`isError\` content shape (tools) or error-JSON contents (resources) without throwing. Dropping the wrapper from any single tool/resource now fails this loop unconditionally — no longer dependent on the CI probe happening to invoke that specific tool.

**Asymmetry comment** — added an inline comment on \`src/index.ts:27\` calling out why \`?? \"\"\` for apiKey vs \`||\` for URL is deliberate (empty key → warn user; empty URL → silently fall back). Pre-empts a future "fix" that would normalize them.

**MISSING_API_KEY_MESSAGE imported in tests** — both \`tools/_shared.test.ts\` and \`resources/_shared.test.ts\` now import the constant rather than hardcoding the substring. A constant rename propagates to assertions automatically.

**Skipped (matches reviewer's non-blocking tag)**
- \`ResourceResult.text\` widening to accept \`blob\` — flagged as TODO-worthy by reviewer; deferring until a real \`blob\` resource shows up.
- Passing whole resource def into \`withMissingApiKeyResource\` to remove URI drift footgun — only two resources today, both colocate the strings; reviewer noted "low risk."
- Runtime image not copying \`package-lock.json\` — reviewer agreed "probably intentional."

**Tests:** 265/265 → **276/276** (+11 from the parameterized loop). CI rerunning on \`196a91f\`.

PR #82, opened 2026-04-27
feat(tools): add output schemas, annotations, and richer descriptions for TDQS

justin-layerv:
## Summary

Improves Glama TDQS (Tool Definition Quality Score) — currently **3.6 / 5** with the rubric flagging missing output schemas, missing behavioral annotations, and thin descriptions across the surface — by addressing every per-tool gap the rubric called out.

## What changes per tool (all 9)

- **`title`** field (separate from `name`) for friendly host display
- **`outputSchema`** declared via Zod and forwarded into `tools/list` (the SDK validates `structuredContent` against it at call time)
- **`annotations`** per the MCP spec — `readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`. Examples:
  - `list_qurls`, `get_qurl` → `readOnlyHint: true`
  - `delete_qurl` → `destructiveHint: true`, `idempotentHint: true`
  - All 9 → `openWorldHint: true` (talks to the qURL API)
- **Descriptions expanded** with: clear one-liner of intent, when-to-use vs sibling tools, return-shape sketch, side-effects, pagination/cursor notes (`list_qurls`)
- **Handlers return `structuredContent`** alongside the existing text content — same payload in both formats

## Server-level

- `src/server.ts` migrated from the deprecated `tool()` to **`registerTool()`** with the config-object pattern. `registerTool` is the path that forwards `outputSchema`/`annotations` into the protocol response; `tool()` would silently drop both.
- New `src/tools/output-schemas.ts` collects Zod schemas mirroring the `client.ts` response shapes — kept in one file so the API spec drift workflow has a single update point.
- `ToolAnnotations` type added to `src/tools/_shared.ts` so factories can type-check their annotation shape without a deep import from the SDK.

## Behavior changes (release-notes worthy)

- **`delete_qurl` is now genuinely idempotent.** Previously, calling `delete_qurl` on an already-revoked or non-existent resource raised the API's 404 to the agent. The handler now swallows the 404 and returns `{ resource_id, revoked: true, was_already_revoked: true, message }` — matching the tool's `annotations.idempotentHint: true` declaration. The new `was_already_revoked` flag in `structuredContent` lets defensive agents distinguish *"I revoked it just now"* from *"the API said 404, treated as no-op"*. Trade-off: the tool can no longer distinguish *"already revoked"* from *"never existed"* by exception alone — both report `revoked: true` (with `was_already_revoked: true`). Other API errors (403, 5xx, etc.) still propagate. Covered by new tests in `delete-qurl.test.ts`.
- **`delete_qurl` confirmation message wording updated.** `"qURL r_xxx has been revoked."` → `"qURL r_xxx is revoked."` — the new wording asserts state rather than action, which is more accurate on the idempotent path. Hosts or transcripts that pattern-match the old string will need a small update.
- **`BatchItemResult` is now a discriminated union on `success`** in `client.ts`, matching the API's mutual-exclusivity contract. Consumers narrowing on `success` get type-safe access to the success-only fields (`qurl_link`, etc.) and the failure-only `error` block. The accompanying `batchItemResultSchema` uses `z.discriminatedUnion("success", ...)` so host UIs generated from the schema render only the relevant branch. Existing callers that didn't narrow continue to work; ones that read both branches at once need to add a discriminator check.

## Additive only on input/text

Input schemas, tool names, and the **shape** of tool responses are unchanged. `structuredContent` is purely additive — older MCP hosts that ignore it see the exact same payload they did before. Goes out as `feat:`, will bump 0.3.1 → 0.4.0 (pre-1.0 minor convention).

## Verified

- `npm run build && npm run lint && npm test` — **366/366 pass** (276 baseline + 90 new tests covering output schemas, annotations, structuredContent round-trip, type-equality, disambiguation guidance, the 404-swallow paths, and `was_already_revoked` flag)
- Rebuilt the Docker image and ran `tools/list` against it: every one of the 9 tools now exposes `title`, `outputSchema`, and `annotations` in the response.

```
9 tools
  ✓ create_qurl               title=True outputSchema=True annotations=True
  ✓ resolve_qurl              title=True outputSchema=True annotations=True
  ✓ list_qurls                title=True outputSchema=True annotations=True
  ✓ get_qurl                  title=True outputSchema=True annotations=True
  ✓ delete_qurl               title=True outputSchema=True annotations=True
  ✓ extend_qurl               title=True outputSchema=True annotations=True
  ✓ update_qurl               title=True outputSchema=True annotations=True
  ✓ mint_link                 title=True outputSchema=True annotations=True
  ✓ batch_create_qurls        title=True outputSchema=True annotations=True
```

## Expected score impact

Glama's pre-PR coherence summary: *"qURL server provides a coherent set of tools for managing secure links. Minor issues include a naming inconsistency with `mint_link` and partial overlap between extend_qurl and update_qurl."* Per-tool: *"Given 9 parameters, no output schema, and no annotations, the description is insufficient"* (list_qurls), *"Complex tool with 8 params and nested objects, but no output schema or behavioral details"* (batch_create_qurls), *"description lacks explicit when/when-not to use this tool vs alternatives"* (delete_qurl, get_qurl).

Tier 1 alone (this PR) targets a lift from **3.6 → ~4.3+**. Naming consistency (`mint_link`, `batch_create_qurls` plural) is intentionally deferred — it's breaking and warrants its own coordinated `feat!:`.

## Test plan

- [x] Build, lint, test green locally (366/366)
- [x] Docker image still passes the full end-to-end probe
- [x] Each tool now exposes `title` + `outputSchema` + `annotations` in `tools/list`
- [x] CI green across CI / CodeQL / Dependency Review / Secrets Scan / Dependency Age Check
- [x] Claude Code Review converged across 12 rounds (LGTM)
- [ ] Glama rescan picks up the new schemas and bumps the score

## Follow-ups (separate issues)

Tracked in #83:
- Verify `delete_qurl` 404 swallow doesn't mask permission denials (cross-tenant 404 vs 403)
- Smoke-test `update_qurl extend_by: \"24h\"` against the live API
- Soften `qurlSchema.status` enum once the API status taxonomy stabilizes
- Audit description verbosity against host UIs (Claude Desktop, Cursor, Glama)
- Verify `update_qurl` clear-field semantics (description: \"\" / tags: [])
- Add `preserve_host` to `client.ts.QURL` and `qurlSchema` (pre-existing drift surfaced by the new strict schemas)

Naming consistency PR (separate, breaking): rename `mint_link` → `mint_qurl_link`, `batch_create_qurls` → `batch_create_qurl` — would land as `feat!:` 0.5.0 with deprecated aliases for one cycle.

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"95baaf8e-5366-4feb-8aa2-bd97466537e2","pid":31627,"procStart":"Mon Apr 27 17:35:40 2026","acquiredAt":1777311573120}
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,1 +1,0 @@
-{"sessionId":"95baaf8e-5366-4feb-8aa2-bd97466537e2","pid":31627,"procStart":"Mon Apr 27 17:35:40 2026","acquiredAt":1777311573120}
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ Thumbs.db
 *.swp
 *.swo
 
+# Claude Code local state
+.claude/
+
 # Environment
 .env
 .env.local
--- a/src/__tests__/helpers.ts
+++ b/src/__tests__/helpers.ts
@@ -1,5 +1,12 @@
 import { vi } from "vitest";
-import type { IQURLClient, QURL, CreateQURLData } from "../client.js";
+import type {
+  BatchCreateOutput,
+  CreateQURLData,
+  IQURLClient,
+  MintLinkOutput,
+  QURL,
+  ResolveOutput,
+} from "../client.js";
 import type { GetPromptResult } from "@modelcontextprotocol/sdk/types.js";
 
 export function makeMockClient(overrides: Partial<IQURLClient> = {}): IQURLClient {
@@ -47,5 +54,49 @@ export function sampleCreateQURLData(overrides: Partial<CreateQURLData> = {}): C
     qurl_site: "https://q_3a7f2c8e91b.qurl.site",
     expires_at: "2026-03-10T00:00:00Z",
     ...overrides,
+  };
+}
+
+export function sampleResolveOutput(overrides: Partial<ResolveOutput> = {}): ResolveOutput {
+  return {
+    target_url: "https://example.com/secret",
+    resource_id: "r_resolved1",
+    access_grant: {
+      expires_in: 300,
+      granted_at: "2026-03-09T12:00:00Z",
+      src_ip: "192.168.1.100",
+    },
+    ...overrides,
+  };
+}
+
+export function sampleMintLinkOutput(overrides: Partial<MintLinkOutput> = {}): MintLinkOutput {
+  return {
+    qurl_link: "https://qurl.link/at_xyz",
+    expires_at: "2026-03-10T00:00:00Z",
+    ...overrides,
+  };
+}
+
+export function sampleBatchCreateOutput(
+  overrides: Partial<BatchCreateOutput["data"]> = {},
+): BatchCreateOutput {
+  return {
+    data: {
+      succeeded: 1,
+      failed: 0,
+      results: [
+        {
+          index: 0,
+          success: true,
+          resource_id: "r_test123",
+          qurl_link: "https://qurl.link/at_abc",
+          qurl_site: "https://q_x.qurl.site",
+          expires_at: "2026-03-10T00:00:00Z",
+        },
+      ],
+      ...overrides,
+    },
+    meta: { request_id: "req_batch" },
   };
 }
--- a/src/__tests__/output-schemas.types.test.ts
+++ b/src/__tests__/output-schemas.types.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expectTypeOf } from "vitest";
 import type { z } from "zod";
 import type {
+  BatchCreateOutput,
   CreateQURLData,
   ListQURLsOutput,
   MintLinkOutput,
@@ -8,6 +9,7 @@ import type {
   ResolveOutput,
 } from "../client.js";
 import {
+  batchCreateOutputSchema,
   createQurlOutputSchema,
   listQurlsOutputSchema,
   mintLinkOutputSchema,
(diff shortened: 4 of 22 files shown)


justin-layerv:
Round-5 follow-ups applied in 1df6d3a:

- `delete_qurl` description now flags the 404-swallow semantic ('Re-deletes return success even if the resource was already revoked or never existed — verify with `get_qurl` first if the ID came from user input.')
- `tdqs-metadata.test.ts` adds a per-tool invariant that `tool.annotations.title === tool.title` so the fields don't silently diverge.

Deferred items tracked in #83:
- 404 swallow vs unauthorized-403 audit against `qurl-service`
- `update_qurl extend_by: "24h"` live-API smoke test
- `qurlSchema.status` enum widening once the API taxonomy stabilizes
- Host-UI description truncation audit (Claude Desktop, Cursor, Glama)

justin-layerv:
Round-7 follow-ups (303f515):

- `delete_qurl` payload message switched from past-tense ("has been revoked") to present-tense ("is revoked") so the idempotent path doesn't claim the call performed the revocation when it may have been a re-delete or a wrong-ID swallow.

Other cr-7 items tracked in #83:
- 404 swallow audit (already item 1 of #83) explicitly covers the cross-tenant case the cr surfaced
- `update_qurl` clear-field semantics (description: "" / tags: []) added as item 6 to #83

Two cr-7 minor notes deliberately not changed:
- `factoryByName` rebuilds tools — keeping the current form (`toolFactories.map((f) => [f(makeMockClient()).name, f])`) because round-4 cr explicitly asked us to drop the index-correlated `tools.map((t, i) => [t.name, toolFactories[i]])` pattern. The two reviewers' preferences pull opposite directions; preserving the correctness-oriented form (no index coupling) over the perf-oriented one in test-only code.
- toStructuredContent `length` constraint comment — defer until a domain object with a legitimate `length` field actually shows up, per the cr's "if/when one shows up" guidance.

justin-layerv:
Round-12 follow-ups:

**Applied:**
- PR description updated to call out the `delete_qurl` confirmation message wording shift ("has been revoked" → "is revoked"), which is technically a string-content change — hosts/transcripts pattern-matching the old wording will need a small update. The original PR body claimed "text content unchanged" without that nuance.

**Tracked in #83:**
- `preserve_host` drift between `api-spec/qurls.yaml`'s ResourceData and `client.ts.QURL` / `qurlSchema`. Pre-existing, made more visible by the new strict schemas. Filed as item 7 of #83.

**Considered, no action:**
- 404 swallow conflation — already documented in description, schema field description, and #83 (item 1). cr is confirming intent; intent confirmed.
- `toStructuredContent` `length?: never` foot-gun — round-6 cr explicitly asked for the array exclusion. Round-12 cr is asking the question rather than directing a change. Keeping the constraint with the existing trade-off comment, which already disclaims that the bound doesn't cover Map/Set/Date/class instances.
- Status enum strictness, resolveQurlOutputSchema error overlap, description verbosity — all explicitly non-blocking per cr; the first is in #83.

PR feels ready; convergence at 12 cr rounds, all four pr-ready criteria green.

Issue #83, opened 2026-04-27
Post-merge follow-ups from PR #82 cr review


justin-layerv:
Tracking deferred follow-ups identified during PR #82's Claude Code Review pass. None blocked the merge; filing here so they don't get lost.

### 1. Verify `delete_qurl` 404 swallow doesn't mask permission denials

**Context.** PR #82 made `delete_qurl` swallow 404 from the API to match its `idempotentHint: true` annotation. Some APIs return 404 to avoid disclosing whether a resource exists for unauthorized callers — if qURL's API does this, an unauthorized caller would now see `revoked: true` instead of a permission error.

**Acceptance.**
- Read `qurl-service`'s handler for `DELETE /v1/qurls/{id}` (and the auth/permission middleware around it) to confirm 404 here is strictly "resource not found", not "not found OR forbidden".
- If it can collapse the two states, either (a) revert the swallow and accept the agent-facing semantic shift, or (b) keep the swallow but add a note to the tool description.

### 2. Smoke-test `update_qurl extend_by: \"24h\"` against the live API

**Context.** The new round-trip test in `tdqs-metadata.test.ts` exercises `update_qurl` with `extend_by: \"24h\"` against a mock client. The tool description claims `update_qurl` accepts `extend_by`, but that hasn't been verified against the live API.

**Acceptance.**
- Run `update_qurl` against staging with `extend_by: \"24h\"` (and `extend_by: \"7d\"` for variety) and confirm the API accepts the duration string syntax. If only `extend_qurl` accepts that format, narrow `update_qurl`'s schema accordingly.

### 3. Soften `qurlSchema.status` enum once API maturity allows

**Context.** `output-schemas.ts` declares `status: z.enum([\"active\", \"revoked\"])` mirroring `client.ts`. If the API ever returns another value (e.g. `\"expired\"`, `\"pending\"`), `structuredContent` validation will hard-fail at call time even though the underlying call succeeded. The API spec drift workflow catches new values at the snapshot level, but not before they reach a host.

**Acceptance.**
- When the API stabilizes its status taxonomy (or as part of the next breaking-change cycle), either: (a) widen the enum to all known values, or (b) replace with `z.string()` + a tighter type-only union, or (c) add `.catch(\"unknown\")` so unknown values pass validation as a sentinel.

### 4. Audit description verbosity against host UIs

[...]

PR #85, opened 2026-04-28
fix(client): add preserve_host to QURL + qurlSchema

justin-layerv:
## Summary

Closes item 7 of #83 — pre-existing drift between the API spec and our types, surfaced by PR #82's new strict-shape Zod schemas.

`api-spec/qurls.yaml:2351-2354` lists `preserve_host` (boolean) on `ResourceData`. It was missing from both `client.ts.QURL` and `qurlSchema` in `output-schemas.ts`. After #82, a server that returns `preserve_host` would have it silently stripped from `structuredContent` (Zod's default strip mode).

## Changes

- `client.ts.QURL` — add `preserve_host?: boolean`
- `output-schemas.ts.qurlSchema` — add the matching `z.boolean().optional()` with a descriptive `.describe()` pointing at the field's semantics (proxy host header preservation, only meaningful with `custom_domain`).

The `expectTypeOf` test in `output-schemas.types.test.ts` now enforces alignment between the two, so future drift on this field will fail compilation.

## Test plan

- [x] `npm run build && npm run lint && npm test` — 366/366 pass
- [ ] CI green
- [ ] Confirm against a `get_qurl` response carrying a `custom_domain` resource that the field reaches us

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -36,6 +36,7 @@ export interface QURL {
   created_at: string;
   status: "active" | "revoked";
   custom_domain?: string | null;
+  preserve_host?: boolean;
   qurl_count?: number;
   qurls?: AccessToken[];
 }
--- a/src/tools/output-schemas.ts
+++ b/src/tools/output-schemas.ts
@@ -72,6 +72,13 @@ export const qurlSchema = z.object({
   // spec-drift workflow catches it before this validation hard-fails.
   status: z.enum(["active", "revoked"]),
   custom_domain: z.string().nullable().optional(),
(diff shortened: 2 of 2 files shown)


justin-layerv:
Good catch on the input-side drift. Tracked as item 8 of #83 — same field pair (`preserve_host` + `custom_domain`) on `UpdateQURLInput`, plus a check whether `preserve_host` is also settable at create time. Scope of this PR is strictly the response-side strict-schema gap from PR #82, so keeping #85 narrow and folding the input-side fix into a separate change.

PR #94, opened 2026-04-29
ci: auto-publish to MCP Registry from release-please workflow

justin-layerv:
## Summary
- Adds a `publish-mcp-registry` job to `release-please.yml`, gated on `release_created` and sequenced after the npm `publish` job.
- Extracts the install/validate/login/publish sequence into a composite action at `.github/actions/publish-mcp-registry/`, used by both the auto-publish (release-please.yml) and manual (publish-mcp-registry.yml) paths — single source of truth for the `mcp-publisher` version + sha256.
- Pins both publish jobs to `ref: ${{ needs.release-please.outputs.tag_name }}` so a post-merge commit on `main` can't sneak into the published artifact.
- Adds a shared `concurrency: mcp-registry-publish` group across both publish jobs (mcp-publisher publish is non-idempotent on the same version).
- Strips the dead `on: push: tags:` trigger from `publish-mcp-registry.yml`; that workflow now only runs via `workflow_dispatch` (manual escape hatch for recovery / registry outage retry).
- Drops the `mcp-publisher --help > /dev/null` sanity-check step — the pinned sha256 already proves byte-for-byte fidelity, and `mcp-publisher validate` would surface any extraction issue immediately.
- Updates `CLAUDE.md` to reflect the new source of truth.

## Why
`release-please` pushes the release tag via `GITHUB_TOKEN`. GitHub suppresses cross-workflow tag triggers from `GITHUB_TOKEN` to prevent recursion, so the standalone workflow's `on: push: tags:` never actually fired on a real release. Every release this round required a manual `gh workflow run "Publish to MCP Registry"` to advance the registry — that's the gap this PR closes.

## Test plan
Neither the new job nor the composite action runs on PR — both gate on `release_created`. CI green here only proves YAML well-formedness; the real exercise happens on the next release-please-created release.

- [ ] After merge, fire `gh workflow run "Publish to MCP Registry (manual)" --repo layervai/qurl-mcp --ref main` once as a smoke test of the composite action via the manual path.
- [ ] Next release-please-created release auto-publishes to both npm and the MCP Registry without manual intervention.


justin-layerv:
Round-2 review addressed:

1. **`timeout-minutes` asymmetry** — added `timeout-minutes: 10` to the npm `publish` job in `release-please.yml`. ✅
3. **Linux/amd64 hardcoded** — added a one-line caller-contract comment in `.github/actions/publish-mcp-registry/action.yml` noting the runner assumption. ✅
5. **CLAUDE.md long paragraph** — split into two bullets (auto path / manual path). ✅

Skipped intentionally:

2. **`set -euo pipefail` on every step** — the other composite-action steps are single-command invocations (`mcp-publisher --help`, `validate`, `login`, `publish`). Default bash exit-on-error semantics for a single command without pipes covers this; `set -euo pipefail` would be no-op boilerplate today. Happy to revisit if `mcp-publisher` ever grows piped output.

4. **Post-merge smoke test** — agree with the framing ("CI green here ≠ proof of fix"). After this merges I'll fire `gh workflow run "Publish to MCP Registry (manual)"` once against `main` to exercise the composite action via the manual path before the next real release relies on it.

justin-layerv:
Round-4 nits addressed:

1. **Sanity-check drop** — added an explicit note in the PR body so a future bisect doesn't second-guess it. ✅
2. **Test-plan checkbox** — rewrote the PR's test plan to set honest expectations: CI green here only proves YAML well-formedness; the new job is exercised on the next release, and a post-merge `workflow_dispatch` smoke test covers the composite action via the manual path before the next release relies on it. ✅
3. **Redundant `if:`** — removed; `needs: publish` inherits release-please's gate via skip-propagation. Updated the comment to call that out. ✅
5. **Pre-PR-94 branch in dispatch UI** — added a one-line comment in `publish-mcp-registry.yml` noting the loud failure at `uses: ./.github/actions/publish-mcp-registry` for branches that don't have the action. ✅

Deferred:

4. **Composite action `inputs.publisher-version` / `inputs.publisher-sha256`** — explicitly skipping per "not worth doing now." Single-source-of-truth in CLAUDE.md is the current contract; if/when we bump from v1.7.2 we can revisit input-parameterization to enable branch-level testing of a publisher bump without editing the action. Worth a follow-up issue if it bites us.

PR #97, opened 2026-04-29
docs(tools): pilot TDQS rewrite for list_qurls and create_qurl

justin-layerv:
## Summary
Glama's TDQS panel currently scores `list_qurls` C (2.9/5) and `create_qurl` B (3.3/5). The same three dimensions drag both: **Behavior**, **Completeness**, and **Usage Guidelines**. `delete_qurl` already scores A (3.5) by structuring its description around four explicit beats — Purpose, When-to-use / When-not-to-use, Behavior, Returns. This PR applies that same template to the two lowest-scoring tools as a calibration pilot before a full pass on the remaining seven.

### What changed
- **`list_qurls`** — locks in the AND-semantics of multi-filter queries, the empty-result-vs-error distinction, default revoked-exclusion, and the omitted-`qurls[]` caveat that pushed disambiguation against `get_qurl`.
- **`create_qurl`** — locks in not-idempotent semantics (twice = two resources), the `qurl_link` one-shot rule, the active-status post-state, and the account-default-expiration trap when `expires_in` is omitted.

### Why pilot, not full pass
Cheapest way to verify Glama's rubric responds to the template. If both tools move into A range on the next rescore, we extend to the other seven (`get_qurl`, `resolve_qurl`, `extend_qurl`, `update_qurl`, `mint_link`, `batch_create_qurls`, `delete_qurl` for consistency) and add a structural lint to `tdqs-metadata.test.ts`.

## Test plan
- [x] `npm run build && npm run lint && npm test` — all 373 tests pass.
- [ ] After merge, wait for Glama rescore (or contact Glama to nudge); confirm `list_qurls` and `create_qurl` move from C/B → A on the TDQS panel.
- [ ] If they don't, re-examine the rubric heuristics before extending the rewrite to the other seven tools.

### DIFF
--- a/src/__tests__/tdqs-metadata.test.ts
+++ b/src/__tests__/tdqs-metadata.test.ts
@@ -1,3 +1,5 @@
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 import { describe, it, expect, vi } from "vitest";
 import type { IQURLClient } from "../client.js";
 import { toolFactories } from "../server.js";
@@ -94,6 +96,19 @@ describe("TDQS tool metadata coverage", () => {
     // structural check — drift inside a nested object surfaces via the
     // round-trip test below.
     //
+    // **One-directional by design.** This enforces description ⊆ schema:
+    // a key claimed in the description must exist in the schema. It does
+    // *not* enforce the inverse (every schema key must be documented).
+    // Adding a schema field without updating the description will pass
+    // this test silently — that's intentional, since not every field is
+    // worth surfacing in the description, but flagging so a future
+    // reader doesn't assume bidirectional coverage.
+    //
+    // **Bare keys only.** The walker matches identifier-followed-by-colon
+    // at depth 0. A future Returns block that wraps keys in backticks or
+    // quotes (e.g. `` `success`: literal `true` ``) will skip them
+    // silently — adapt the walker if/when that pattern shows up.
+    //
     // Tools without a Returns block (or whose block describes a
     // discriminated/wrapped shape) are skipped here.
     const topLevelKeys = (body: string): string[] => {
@@ -140,6 +155,33 @@ describe("TDQS tool metadata coverage", () => {
         }
       });
     }
+  });
+
+  describe("description claims pinned against api-spec defaults", () => {
+    // Strong factual claims about API defaults (e.g. "expires_in defaults
+    // to 24h") become silent lies if the server-side default moves. Read
+    // the snapshot at api-spec/qurls.yaml — the same file the spec-drift
+    // workflow already monitors — and assert the description quotes the
+    // value verbatim. If the spec changes, the drift workflow updates the
+    // file and this test forces a description update at the same time.
+    const specPath = fileURLToPath(new URL("../../api-spec/qurls.yaml", import.meta.url));
+    const spec = readFileSync(specPath, "utf8");
+
+    it("create_qurl description matches the spec's expires_in default", () => {
+      // Anchor on the schema block (`expires_in:` followed by `type:`),
+      // not on the first `expires_in:` token in the file — earlier
+      // occurrences are example values inside response samples and
+      // would let an unrelated `Default:` field between here and the
+      // real schema slip into the capture group.
+      const match = spec.match(/expires_in:\s*\n\s*type:[\s\S]*?Default:\s*(\d+\s*[a-z]+)/);
+      expect(match, "expected to find an `expires_in: type: … Default: …` block in api-spec/qurls.yaml").not.toBeNull();
+      const specDefault = match![1].trim();
+      const description = tools.find((t) => t.name === "create_qurl")!.description;
+      expect(
+        description,
+        `create_qurl description must mention the spec's expires_in default '${specDefault}'`,
+      ).toContain(specDefault);
+    });
   });
 
   describe("safety hints match tool semantics", () => {
--- a/src/__tests__/tdqs-metadata.test.ts
+++ b/src/__tests__/tdqs-metadata.test.ts
@@ -171,11 +171,13 @@ describe("TDQS tool metadata coverage", () => {
     const spec = readFileSync(specPath, "utf8");
 
     it("create_qurl description matches the spec's expires_in default", () => {
-      // The spec has two `expires_in:` properties (one each on the
-      // CreateQurl and UpdateQurl request schemas) — only the create-side
-      // declares a `Default:`. Iterate every match and pick the block
-      // that carries the default; relying on source order would silently
-      // pick the wrong block if the schemas ever swap positions.
+      // The spec has two `expires_in: type: string` blocks (CreateQurlRequest
+      // and MintLinkRequest) — only the create-side declares a `Default:`
+      // inside its own description (MintLink puts the 24h default in the
+      // parent schema description, not the property's). Iterate every match
+      // and pick the block that carries `Default:`; relying on source
+      // order would silently pick the wrong block if the schemas ever
+      // swap positions.
       //
       // Indent matcher is `[ ]{12,}` (description-body depth) rather than
       // `\s+` so the captured block stops at the next outdented line —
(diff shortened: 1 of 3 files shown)


justin-layerv:
Round-3 cr items addressed in 8c13ac2:

- 🟡 **Default: matcher scope** — switched to a two-step extraction (capture expires_in's `description: |` block, then search Default within only that block). Prior regex could have walked past expires_in into a sibling property's Default line.
- 🟢 **`RFC3339` → `RFC 3339`** — aligned create_qurl Returns block with the existing client.ts/input-schema convention.
- 🟢 **"Filters AND together" claim** — filed as #98 (documentation, priority: low) to be locked down via spec docs or a staging-gated behavioral test in a follow-up. The current claim matches REST/OpenAPI convention but isn't contractual from the spec snapshot.

Items left as informational pending Glama rescore (per cr's "worth confirming on the rescore before assuming the template is the variable"):

- One-directional schema check gap for list_qurls's `get_qurl`-shape reference. If TDQS rewards explicit field listings, the implicit reference may keep the rubric below A even after this rewrite — the rescore is the cheapest way to find out before the template extends to the other seven tools.
- Test naming `Returns block keys exist in outputSchema` reads as a generic invariant; once the pilot extends, will hoist the walker to a shared helper or rename.
- `(one-time)` annotation on `qurl_link` was flagged "fine as-is, just noting redundancy" — leaving as-is per the explicit guidance.

justin-layerv:
Round-4 cr items addressed in ffd193c:

**Substantive ("Approve when"):**
- ✅ #1 — explicit `By default only active qURLs are returned` restored to list_qurls description, mirroring delete_qurl's existing claim. Spec pinning tracked in #99.
- ✅ #2 — softened default-sort claim with inline caveat that only the `desc` direction is spec-pinned (api-spec/qurls.yaml:295-302). Field default tracked in #99.

**Other items:**
- ✅ #3 — `(via the `q` parameter)` named in list_qurls description.
- ✅ #4 — `(one-time)` → `(shown once)` on qurl_link in create_qurl Returns block.
- ✅ #6 — `depth = Math.max(0, depth - 1)` guard on topLevelKeys walker; identifier fast-forward optimization left out as the agent noted it's negligible at <2KB description sizes.
- ✅ #5 — improved the spec-pin failure message to name the specific structural assumptions (immediate `type: string`, then `description: |`); declined adding js-yaml just for one call site, per round-2 simplify guidance.
- 📋 #7 (coverage asymmetry) — list_qurls has more pinnable facts (limit ceiling, default page size). Not adding in this pilot pending Glama rescore confirming the template direction; will extend the spec-pinning mechanism in the full-pass PR.
- 📋 #8 (description length budget) — flagged for the full-pass PR, not actionable here.

All four pr-ready criteria currently green: would-a-DE-accept (yes pending this round-4), CI passing, all cr items addressed including nits, /simplify clean.

justin-layerv:
Round-6 cr items addressed in fd76d11:

- ✅ **#1 (Lazy regex truncation)** — replaced `\{([\s\S]+?)\}` with a brace-balanced extractor that walks forward from the marker's opening `{` and returns the slice between the outermost matching pair. The lint is now nesting-correct, not just nesting-correct-by-luck. Verified all 378 tests still pass.

Items deferred (acknowledged but out of scope for the pilot):

- 📋 **#2 (One-directional design)** — adding the inverse direction requires defining an opt-in 'fields-worth-documenting' list per tool. Will fold in alongside the full-pass PR where that list naturally takes shape across all 9 tools.
- 📋 **#3 (Plan-tier max caps in create_qurl)** — agree it's out of scope for the pilot; will revisit when extending. Captured as a note in the full-pass plan.
- 📋 **#5 (Sort-field default)** — already tracked in #99. When the spec pins the field default, will mirror the `expires_in` assertion to lock `created_at` against the spec the same way.
- 📋 **#6 (`readFileSync` at module-load)** — agree the failure-scoping benefit is small; defer.

Observation acknowledged:

- 📋 **#4 (Token cost)** — explicit Glama tradeoff; flagging in the full-pass PR's design doc.

Ready for re-review. All four pr-ready criteria currently green pending this round.

justin-layerv:
Round-7 cr items addressed in 8c3f30e:

**Actionable (applied):**
- ✅ **Nit (status .describe alignment)** — schema-level `status` describe now mentions `Defaults to 'active' when omitted` so the host argument hint matches the tool description.
- ✅ **#3 (topLevelKeys prose-colon hardening)** — added string/backtick-literal tracking to the walker, with escape-skip. A future Returns block with a quoted prose value containing a colon would have false-positive captured the preceding word as a key. Walker now skips literals correctly.

**Deferred:**
- 📋 **#1 (default-active claim softening)** — round-4 simplify explicitly trimmed the epistemics caveat I'd added in ffd193c on grounds that overhedging in LLM-facing prose invites defensive over-specification. Pinning is tracked in #99 (preferred fix is documenting the default in the spec, not narrating uncertainty in the description).
- 📋 **#2 (broaden spec-pin to Record<string, string>)** — your comment explicitly notes "worth doing before extending the template to the remaining seven tools." Will fold into the full-pass PR where the additional fields-to-pin (`one_time_use`, etc.) take shape across all 9 tools simultaneously.
- 📋 **#4 (token-cost retro)** — observation; not actionable until the Glama rescore reports back.

Ready for re-review.

justin-layerv:
Round-8 cr items addressed in e828156:

**Applied:**
- ✅ **list_qurls TODO(#99) marker** — added inline above the unverified default-active and default-sort-field claims so a future grep ties them to the tracking issue. Descriptions themselves unchanged (overhedging was deliberately trimmed in round 5).
- ✅ **Returns-block test rename** — `lists only schema keys` → `claims only keys present in schema` to match the actual description ⊆ schema direction.
- ✅ **Round-trip-test comment correction** — fixed the load-bearing comment to accurately describe what the round-trip test catches (handler ↔ schema, not description's nested claims).
- ✅ **`**24h**` bold-marker tightening** — the spec-pin assertion now requires the Behavior-beat bolded form, closing the Example-line loophole. If the spec moves to e.g. 48h while the Example still says '24h', the test now fails.

**Deferred:**
- 📋 **mint_link 24h asymmetric coverage** — folded into the full-pass extension PR (when the template hits all 9 tools, the spec-pin matcher will also extend to mint_link's parent-schema-default form per round-7 cr's #2 suggestion).
- 📋 **"more cheaply" prose claim on get_qurl** — cr noted it's prose, not a measured fact; no action requested. Left as agent-routing guidance.

Ready for re-review.

justin-layerv:
Round-9 cr items addressed in 09262e8:

**Applied:**
- ✅ **Sentinel tightening** — switched the alignment guard from a generic `active` substring to claim-specific phrases (`Defaults to 'active'` on the schema describe side, `By default only \`active\`` on the description side). The test now actually fails for the drift mode it was designed to catch.

**Deferred (forward-looking observations from cr):**
- 📋 **Schema describe ↔ description redundancy** — for the full-pass PR. cr suggests deciding which is canonical and trimming the other when extending; doing it now would couple pilot scope to extension.
- 📋 **`tools.find()` vs hoisted `byName` Map** — cosmetic, flagged in prior rounds. Will fold into the test-helper hoist when the walker extends to all 9 tools.
- 📋 **Description token cost** — explicit Glama tradeoff; post-rescore retro will weigh this against the rubric movement before committing the full pass.

Ready for re-review.

justin-layerv:
Round-10 cr items addressed in e83d582:

**Applied:**
- ✅ **Idempotency nuance** — added `(this tool doesn't surface the underlying API's \`Idempotency-Key\` header)` to the create_qurl Behavior beat. Cheap clarification; an agent reading both the MCP description and the broader API spec no longer needs to cross-reference to learn the gap.

**Confirmed (no code change needed):**
- ✅ **TODO(#99) link** — confirming #99 exists at https://github.com/layervai/qurl-mcp/issues/99 ("Lock down list_qurls default-behavior claims", labels: documentation/priority: low). The `TODO(#99):` shorthand resolves correctly in GitHub UI; nothing further needed.

**Deferred to full-pass PR:**
- 📋 **Silent-skip guard for tools without Returns block** — useful only when the template extends; will add `expect(toolsWithReturnsBlock.length).toBeGreaterThanOrEqual(N)` when the count moves from 2/9 to 9/9.
- 📋 **`expires_in.describe()` sentinel pin** — per cr: "worth adding while extending." Will fold into the full-pass schema-describe alignment work.
- 📋 **Token cost retro** — explicit Glama tradeoff measurement; gated on the rescore landing first.

Ready for re-review.

PR #100, opened 2026-04-29
fix(tools): close items 1 and 3 from issue #83 follow-ups

justin-layerv:
## Summary
Closes the two #83 items that don't require live API/staging access. Items 5/7/8 were already done in prior PRs; items 2/4/6 stay open pending staging.

### Item 1 — delete_qurl 404 swallow vs permission denial
**Verified** against qurl-service/internal/service/qurl_service.go:407-409:
```go
if resource.OwnerID != ownerID {
    return domain.ErrResourceNotFound
}
```
The service deliberately collapses ownership-mismatch into ErrResourceNotFound to avoid existence disclosure. The MCP tool's 404 swallow honors the idempotent contract but does mask permission errors as a side effect.

**Fix (option (b) from the issue):** keep the swallow + document the conflation in the tool description so agents that need to confirm ownership know to `get_qurl` first. Handler comment cites the exact source line of the collapse.

### Item 3 — soften qurlSchema.status enum
**Fix (option (c) from the issue):** widen the enum to include sentinel `"unknown"` and `.catch("unknown")` so unanticipated API values (e.g. `"expired"`, `"pending"`) don't hard-fail structuredContent validation between weekly drift-detection runs. Mirror into client.ts QURL.status to keep the type-equality test passing.

Two new round-trip tests in output-schemas.types.test.ts lock the behavior in.

## Items NOT in scope
- **Item 2** — smoke-test update_qurl extend_by (needs staging API access)
- **Item 4** — audit description rendering on Claude Desktop / Cursor / Glama (needs host UI access)
- **Item 6** — verify update_qurl clear-field semantics (needs staging API access)

These remain tracked in #83 with their original context.

## Test plan
- [x] `npm run build && npm run lint && npm test` — all 381 tests pass (379 + 2 new .catch tests).
- [x] `output-schemas.types.test.ts` `expectTypeOf<z.infer<typeof qurlSchema>>().toEqualTypeOf<QURL>()` still passes after widening both sides.
- [ ] On merge, comment on #83 marking items 1, 3 as done.

### DIFF
--- a/src/__tests__/helpers.ts
+++ b/src/__tests__/helpers.ts
@@ -33,6 +33,11 @@ export function getPromptText(result: GetPromptResult, index = 0): string {
   return content.text;
 }
 
+// Defaults satisfy the AccessToken type but aren't behaviorally
+// coherent (e.g. `max_sessions: 0` with `one_time_use: false` is a
+// shape that grants nothing). Suitable for schema/structural tests;
+// override fields explicitly when exercising session-accounting or
+// expiration logic.
 export function sampleAccessToken(overrides: Partial<AccessToken> = {}): AccessToken {
   return {
     qurl_id: "q_abc123",
--- a/src/__tests__/helpers.ts
+++ b/src/__tests__/helpers.ts
@@ -1,5 +1,6 @@
 import { vi } from "vitest";
 import type {
+  AccessToken,
   BatchCreateOutput,
   CreateQURLData,
   IQURLClient,
@@ -30,6 +31,20 @@ export function getPromptText(result: GetPromptResult, index = 0): string {
   if (typeof content === "string") return content;
   if (content.type !== "text") throw new Error(`Expected text content, got ${content.type}`);
   return content.text;
+}
+
+export function sampleAccessToken(overrides: Partial<AccessToken> = {}): AccessToken {
+  return {
+    qurl_id: "q_abc123",
+    status: "active",
+    one_time_use: false,
+    max_sessions: 0,
+    session_duration: 3600,
+    use_count: 0,
+    created_at: "2026-01-01T00:00:00Z",
+    expires_at: "2026-12-31T00:00:00Z",
+    ...overrides,
+  };
 }
 
 export function sampleQURL(overrides: Partial<QURL> = {}): QURL {
--- a/src/__tests__/output-schemas.types.test.ts
+++ b/src/__tests__/output-schemas.types.test.ts
@@ -109,9 +109,9 @@ describe("qurlSchema.status drift tolerance", () => {
         status: 42,
       }).status,
     ).toBe("unknown");
-    const { status: _drop, ...withoutStatus } = sampleQURL();
-    void _drop;
-    expect(qurlSchema.parse(withoutStatus as Record<string, unknown>).status).toBe("unknown");
+    const withoutStatus: Record<string, unknown> = { ...sampleQURL() };
+    delete withoutStatus.status;
+    expect(qurlSchema.parse(withoutStatus).status).toBe("unknown");
   });
 
   it("accepts the documented access-token statuses on the nested array", () => {
--- a/src/__tests__/output-schemas.types.test.ts
+++ b/src/__tests__/output-schemas.types.test.ts
@@ -146,4 +146,22 @@ describe("qurlSchema.status drift tolerance", () => {
     expect(parsed.qurls).toHaveLength(1);
     expect(parsed.qurls?.[0].status).toBe("unknown");
   });
+
+  it("survives schema composition through listQurlsOutputSchema", () => {
+    // qurlSchema is referenced (not inlined) by listQurlsOutputSchema, so
+    // .catch() should propagate through composition. Guard against a
+    // future refactor that inlines or re-shapes the field at the list
+    // boundary.
+    const parsed = listQurlsOutputSchema.parse({
+      data: [
+        {
+          ...sampleQURL(),
+          // @ts-expect-error simulating an out-of-spec API value through the list envelope
+          status: "pending",
+        },
+      ],
+      meta: { has_more: false },
+    });
+    expect(parsed.data[0].status).toBe("unknown");
+  });
 });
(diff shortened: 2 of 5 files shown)


justin-layerv:
Round-1 cr items addressed in 0e656a3:

**Applied:**
- ✅ **#1 (accessTokenSchema parallel test)** — added a nested round-trip test that parses a sampleQURL whose `qurls[0].status` carries an out-of-spec value, asserting the .catch coercion fires through the nested schema. A future refactor that strips .catch from accessTokenSchema now fails this test.

**Deferred (with tracking issue):**
- 📋 **#2 (silent coercion / operator visibility)** — filed as #101. Behavior change worth tracking separately; preferred fix is a custom catch handler that logs to stderr (same path as the existing `QURL_API_KEY is not set` warning).

**Acknowledged, no action:**
- 📋 **#3 (`unknown` collision)** — comment in client.ts captures the tradeoff explicitly. cr said "Your call — the comment captures the tradeoff."
- 📋 **Description length** — already deferred to #83 item 4 (host UI rendering audit).

Ready for re-review.

justin-layerv:
Round-3 cr addressed in b978cdb:

**Applied:**
- ✅ **Shape-regression scope** — added the suggested comment addendum to output-schemas.ts acknowledging that `.catch()` triggers on null / wrong-type / missing values too, not just unrecognized enum strings. References #101 for operator-visibility follow-up.
- ✅ **Shape-regression test** — added a test that hits all three non-string failure modes (`null`, `42`, missing field) and asserts each coerces to `unknown`. Uses `@ts-expect-error` for each boundary violation per the established pattern. A future refactor that narrows `.catch()` to strings-only now trips this test.

Test count: 382 → 383.

cr's "Approve" verdict was non-blocking on this; folding it in proactively to keep the doc and behavior aligned for the next maintainer.

justin-layerv:
Round-5 cr addressed in 8e58b22:

**Applied (all 4):**
- ✅ **#1 (`expired` first-class)** — verified the spec inconsistency: api-spec/qurls.yaml:2742 lists `enum: [active, revoked]` but lines 2745-2746 say resources past expires_at are reported as `\"expired\"`. Widened qurlSchema.status to `['active', 'revoked', 'expired', 'unknown']` and mirrored in client.ts QURL.status. `.catch('unknown')` still safety-nets truly new values; `expired` round-trips with semantics intact. Updated the unrecognized-value test to use `pending` and added an `expired` pass-through assertion.
- ✅ **#2 (get_qurl 404 ambiguity)** — verified against qurl-service: GetQurl also collapses ownership-mismatch into ErrResourceNotFound. Updated description to clarify "200 from get_qurl confirms ownership; 404 doesn't distinguish 'not yours' from 'doesn't exist'". Handler comment now references both RevokeQurl and GetQurl with the symmetry called out.
- ✅ **#3 (nested-token pass-through test)** — iterated all four spec-documented access-token statuses (active/consumed/expired/revoked) on a nested fixture. Cheap insurance against an enum typo on accessTokenSchema.
- ✅ **#4 (cast cleanup)** — `Parameters<typeof qurlSchema.parse>[0]` → `Record<string, unknown>`. Same effective type, clearer intent.

Test count: 383 → 384.

justin-layerv:
Round-7 cr addressed in dfd1a15:

**Applied:**
- ✅ **#3 (qurlSchema.status comment density)** — trimmed from 17 lines to 7, pointing at the test file as the source of truth for fail-soft scope.
- ✅ **#4 (sampleAccessToken defaults synthetic)** — added a comment noting the defaults are shape-only and not behaviorally coherent.

**Filed as follow-up:**
- 📋 **#1 (list_qurls filter accepts `expired`?)** — filed as #102. Description-only follow-up gated on staging verification, per cr's "reasonable to defer to a separate PR."

**Acted on out-of-band:**
- 📋 **#2 (operator-visibility prioritization)** — bumped #101 from `priority: low` → `priority: medium` so the silent-fail-soft posture doesn't sit unobserved indefinitely.

Test count unchanged (384 → 384). All criteria green pending merge.

justin-layerv:
Round-10 cr addressed:

- **Suggestion 1** (missing-field telemetry signal): captured in #101 — added a "Signals to distinguish" section that splits unrecognized-enum / missing-field / real-`"unknown"`-from-wire as distinct triage cases.
- **Suggestion 2** (`.catch()` callback as the disambiguation point): also captured in the same #101 section — noted that Zod 4's `.catch()` callback receives `ctx.input`, so the collision is observable inside the schema callback rather than downstream.
- **Suggestion 3** (sampleAccessToken default coherence): no change — cr says "fine as-is" given the existing comment.
- **Suggestion 4** (delete-qurl wording: "actual revoke" loses a nuance): tightened to "a successful revoke on this call" in `d52330c`. Captures both the action and the temporal scoping that `was_already_revoked: false` actually conveys.

justin-layerv:
Round-11 cr addressed in `745d0bc`:

- **Suggestion 1** (`"expired"` is documentation-driven not enum-driven): filed as #103 — proposes either (a) widening the qurl-service spec's `enum:` line so the next drift run lands the change, or (b) demoting `"expired"` here to a drift sentinel. (a) is preferred since the value is observably emitted today.
- **Suggestion 2** (`.catch()` masks malformed payloads — telemetry must log raw input): added a top-level "Required: capture raw `ctx.input`" section to #101 making this explicit. The signals-to-distinguish section already covered the three triage cases (unknown-enum / missing-field / real-`"unknown"`-from-wire), but the raw-input requirement is now called out as a hard requirement on whichever shape #101 takes.
- **Suggestion 3** (nested-schema coverage via `listQurlsOutputSchema`): added the round-trip test in `output-schemas.types.test.ts`. Drifted `status` → coerced to `"unknown"` → confirmed through the list envelope. Guards against a future refactor that inlines or re-shapes `qurlSchema` at the list boundary.
- **Suggestion 4** (`delete_qurl` line 31 wording): tightened to "call `get_qurl` first — a 200 confirms ownership; a thrown 404 is equally ambiguous on that endpoint too." Clarifies that `get_qurl`'s 404 is a thrown `QURLAPIError`, not a 404 in the response shape.
- **Suggestion 5** (sampleAccessToken defaults): no change — cr says "fine as-is."

385 tests pass (added 1), lint + build clean.

PR #140, opened 2026-05-31
ci: bump issue-priority reusable shim to ops-routines-workflows v0.7.0

justin-layerv:
## What

Bumps the `issue-priority` reusable shim to `ops-routines-workflows` **v0.7.0** (ref-only — no other changes).

## Why

Previously, an issue created **out-of-band** (REST API / `gh issue create` / a stripped label) with no `priority: *` label hit the reusable's passive fall-through: a failed run that gates nothing, so the issue persisted with no priority. The required-dropdown issue forms only cover the web-UI path.

GitHub has no pre-create hook for issues, so v0.7.0 makes the gap **self-healing** instead: it attaches a `needs: priority` holding label (auto-created if the repo doesn't define it) + a one-time comment, keeps the run red until a real priority is set, then **strips the holding label automatically** once one is. No silent default — the author defines the priority. Bot-filed issues stay exempt at creation to avoid label-event recursion.

## Safety

Ref-only bump; the reusable's interface (`issues: write`, no inputs) is unchanged, it's fully tested upstream, and the behavior was validated live before this rollout. Released as `ops-routines-workflows` v0.7.0.

### DIFF
--- a/.github/workflows/issue-priority.yml
+++ b/.github/workflows/issue-priority.yml
@@ -48,6 +48,6 @@ jobs:
     # reusable's own `sender.type != 'Bot'` check short-circuits them
     # before setFailed — net effect is a near-instant no-op run.
     if: ${{ github.actor != 'github-actions[bot]' }}
-    uses: layervai/ops-routines-workflows/.github/workflows/issue-priority.yml@ca41c7123c507ad45331a723851da6c010541154 # v0.6.2
+    uses: layervai/ops-routines-workflows/.github/workflows/issue-priority.yml@f9a4ab4e4697980ebd21efed7d643441f0ea0993 # v0.7.0
     permissions:
       issues: write


justin-layerv:
Thanks for the review. The change is identical across the fleet bump, so addressing the recurring notes uniformly:

**SHA ↔ tag (the trust anchor).** Confirmed the pin is the tagged release — `v0.7.0 → f9a4ab4e4697980ebd21efed7d643441f0ea0993`:
```
$ gh api repos/layervai/ops-routines-workflows/git/refs/tags/v0.7.0 --jq '{tag:.ref,sha:.object.sha}'
{"sha":"f9a4ab4e4697980ebd21efed7d643441f0ea0993","tag":"refs/tags/v0.7.0"}
```
Matches the `# v0.7.0` comment exactly.

**Concurrency comment / `cancel-in-progress: false`.** Still accurate, not stale. v0.7.0 performs every *label* mutation as a single atomic `setLabels` PUT (add-holding, strip-holding, form-swap), so the comment's load-bearing claim holds and the "non-atomic remove→add leaves an issue with no priority label" scenario remains genuinely hypothetical. What v0.7.0 adds is an idempotent `createLabel` and a dedup-guarded one-time `createComment` — both convergent under a mid-flight cancel: the next run re-evaluates from the issue's current labels (holding-present → skip comment; comment-posted-but-label-missing → re-apply), so a cancelled run still can't leave an issue priority-less. So `cancel-in-progress: false` stays correctly justified. (Happy to enrich the comment to mention the new side effects, but it isn't inaccurate.)

**Age-check gate.** Passed — the `ops-routines` age-check reusable skips self-references (`AGE_CHECK_SKIP_SELF_REFS` includes `layervai/ops-routines-workflows`), so the minutes-old v0.7.0 SHA isn't gated despite `min_age_days: 7`.

**Holding-label auto-creation.** Validated live on nhp before this sweep: a throwaway out-of-band issue auto-provisioned `needs: priority`, got the comment + a red run, then self-healed (label stripped, run green) once a priority was added. No label-not-found error.

PR #157, opened 2026-06-13
feat(client): use the @layervai/qurl SDK instead of the hand-rolled client

justin-layerv:
## What

Replaces the hand-rolled `src/client.ts` (its own `fetch`/retry/RFC-7807 parsing) with a thin **adapter over the published [`@layervai/qurl`](https://www.npmjs.com/package/@layervai/qurl) SDK**. This is the qurl-mcp half of the "use the layervai TypeScript client instead of rolling our own" migration — the SDK shipped to npm as `@layervai/qurl@0.2.0` (built from `layervai/qurl-typescript`).

The SDK now owns all transport: fetch, retries, error parsing, response shape validation. `client.ts` keeps the exact same public surface — every domain type, the `IQURLClient` interface, `QURLAPIError`, and `MISSING_API_KEY_MESSAGE` are unchanged — and only **translates** between the SDK and that contract:

| Concern | Translation |
|---|---|
| Return envelopes | Re-wrap the SDK's unwrapped returns in `{ data }` |
| Field name | Rename the SDK's `access_tokens` → `qurls` |
| List / batch / sessions | Reshape `{qurls,next_cursor,has_more}`, flatten `BatchCreateOutput`, map `{sessions}`→`{data}` and `{terminated}`→`{data:{terminated}}` |
| Method names | `revokeQurlToken`→`revokeResourceQurl`, `updateQurlToken`→`updateResourceQurl` |
| Errors | `QURLError` subclasses → `QURLAPIError`, preserving `statusCode`/`code`/`requestId` (delete-qurl branches on `statusCode === 404`) |
| Keyless boot | SDK built **lazily** on first call (its constructor throws on an empty key); `index.ts` still boots keyless for MCP introspection, and the missing-key path still emits `MISSING_API_KEY_MESSAGE` |

## Why it's low-risk

The `IQURLClient` interface and all output shapes are identical, so the **24 tool/resource/output-schema test files pass untouched** — direct evidence the MCP's external contract (tool I/O, `qurls`, `{data}` wrapping) is preserved. Only `client.test.ts` is rewritten: it previously tested the hand-rolled HTTP layer (now the SDK's job) and now covers the adapter's translation logic (keyless boot, `{data}` wrapping, `access_tokens`→`qurls`, envelope reshaping, method renames, `QURLError`→`QURLAPIError`).

## Verification (local)

- `npm run build` ✅ · `npm run lint` ✅ · `npm run format:check` ✅
- `npm test` ✅ — **421 passed (25 files)**
- `npm ci` ✅ — resolves `@layervai/qurl@0.2.0` from the lockfile (the repo's `min-release-age=7` gate applies to `npm install` resolution, not `npm ci`, so CI is unaffected even though the SDK was just published)

## Notes / follow-ups

- **Minor parity gap:** the SDK's `QURLError` does not surface `tombstone` (410 path), so `QURLAPIError.tombstone` is now always `undefined`. No MCP tool reads it, so there's no functional impact — flagging for an optional future SDK enhancement.
- A local `npm install` of `@layervai/qurl` is age-gated until ~2026-06-20 (the 7-day supply-chain window); use `--min-release-age=0` for local lockfile regen until then. CI (`npm ci`) is unaffected.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -56,7 +56,7 @@ qurl-mcp/
 ├── src/
 │   ├── index.ts           # Entry point, env validation, stdio transport
 │   ├── server.ts          # MCP server factory, tool/resource/prompt registration
-│   ├── client.ts          # TypeScript qURL API client
+│   ├── client.ts          # Adapter over the @layervai/qurl SDK (IQURLClient + QURLAPIError)
 │   ├── tools/
 │   │   ├── _shared.ts       # resourceIdSchema, zodErrorToToolResult
 │   │   ├── create-qurl.ts
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.4.0",
       "license": "MIT",
       "dependencies": {
+        "@layervai/qurl": "^0.2.0",
         "@modelcontextprotocol/sdk": "^1.12.1",
         "zod": "^4.4.3"
       },
@@ -253,6 +254,15 @@
       "integrity": "sha512-cYQ9310grqxueWbl+WuIUIaiUaDcj7WOq5fVhEljNVgRfOUhY9fy2zTvfoqWsnebh8Sl70VScFbICvJnLKB0Og==",
       "dev": true,
       "license": "MIT"
+    },
+    "node_modules/@layervai/qurl": {
+      "version": "0.2.0",
+      "resolved": "https://registry.npmjs.org/@layervai/qurl/-/qurl-0.2.0.tgz",
+      "integrity": "sha512-+ldlWnnXPBQKAkGO/5v00FdK6bOvK2kMW7Cii1/BO4RG5Vu7SiBinvt37YcgmNJtkQY+bw3kw127q5o2Hij0CA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=20"
+      }
     },
     "node_modules/@modelcontextprotocol/sdk": {
       "version": "1.29.0",
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "url": "https://github.com/layervai/qurl-mcp"
   },
   "dependencies": {
+    "@layervai/qurl": "^0.2.0",
     "@modelcontextprotocol/sdk": "^1.12.1",
     "zod": "^4.4.3"
   },
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "vitest": "^4.1.5"
   },
   "engines": {
-    "node": ">=18"
+    "node": ">=20"
   },
   "files": [
     "dist/"
--- a/src/__tests__/client.test.ts
+++ b/src/__tests__/client.test.ts
@@ -1,977 +1,227 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { QURLClient, QURLAPIError } from "../client.js";
 
-function stubFetch(body: unknown, status = 200) {
-  const mock = vi.fn().mockResolvedValue({
-    ok: status >= 200 && status < 300,
-    status,
-    text: () => Promise.resolve(JSON.stringify(body)),
+// `client.ts` is now a thin adapter over the `@layervai/qurl` SDK. These tests
+// mock the SDK so we can assert the adapter's translation layer in isolation:
+// lazy/keyless construction, `{ data }` re-wrapping, the `access_tokens` →
+// `qurls` rename, the list/batch/session envelope reshaping, the method
+// renames, and the `QURLError` → `QURLAPIError` mapping. The HTTP/retry/parse
+// behavior the old hand-rolled client used to be tested for now lives in (and
+// is tested by) the SDK itself.
+
+const { sdk, SDKClientMock } = vi.hoisted(() => {
+  const sdk = {
+    create: vi.fn(),
+    get: vi.fn(),
+    list: vi.fn(),
+    delete: vi.fn(),
+    update: vi.fn(),
+    updateResource: vi.fn(),
+    extend: vi.fn(),
+    resolve: vi.fn(),
+    getQuota: vi.fn(),
+    mintLink: vi.fn(),
+    batchCreate: vi.fn(),
+    revokeResourceQurl: vi.fn(),
+    updateResourceQurl: vi.fn(),
+    listResourceSessions: vi.fn(),
+    terminateResourceSession: vi.fn(),
+    terminateAllResourceSessions: vi.fn(),
+  };
+  // The SDK constructor returns our shared stub instance. Must be a regular
+  // function (not an arrow) so the adapter's `new SDKQURLClient(...)` works —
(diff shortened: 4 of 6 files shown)


justin-layerv:
Thanks — addressed the review in `bf47450`:

**High — batchCreate HTTP 400 passthrough:** Verified there's **no regression**. The installed SDK has `BATCH_PASSTHROUGH_STATUSES = [400]`, so `sdk.batchCreate` passes through the all-items-failed 400 body as a populated `BatchCreateOutput` instead of throwing — exactly the behavior `batch-create.ts` depends on. The adapter just reshapes it. **Added a test** for the all-failed case (`succeeded: 0, failed: 2, results: [...]`) plus a call-site comment, so the behavior is now locked.

**Medium — `as unknown as` erasing type safety:** Tightened. Dropped the double-casts: **8 of 9 inputs are structurally assignable to the SDK input types and now pass through unannotated**, so a future SDK input-field rename/removal fails to compile. Outputs use a single `as` (gross drift still rejected). Only `updateResourceQurl` keeps a bridge cast — the SDK types that input as an `extend_by`-XOR-`expires_at` discriminated union, while the MCP allows both and enforces the XOR via a zod refinement.

**Behavioral — `extendQURL` changed path:** `sdk.extend` is implemented in the SDK as the same `PATCH /v1/qurls/{id}` (it delegates to `update`), so the endpoint/semantics are unchanged. **Added a dedicated `extendQURL` translation test.**

**Minor:**
- `mapResource` now drops the `qurls` key entirely when the SDK omits `access_tokens` (no more `qurls: undefined`).
- `translateError` uses the SDK's always-populated `QURLError.detail` (which is `detail ?? title`), not `detail || message` — the `title` rung is preserved and there's no empty-string edge case.
- `listQURLs`' omission of `meta.page_size`/`request_id` is now commented as intentional (the SDK's list envelope doesn't surface them).
- `tombstone` gap: confirmed inert (no `src/` consumer); the cast comment now states plainly the SDK error type doesn't expose it.
- delete-qurl 404: the SDK's `parseError` falls back to `{ status: <httpStatus>, code: "unknown" }` on a non-JSON body, so a 404 with an HTML proxy page still maps to `NotFoundError` with status 404 → `QURLAPIError.statusCode === 404` → the idempotent swallow fires.

**On the `age-check` "blocking" item:** this was a deliberate, owner-authorized decision, not an oversight. `@layervai/qurl` is first-party, published via our own release process, so the `age-check-bypass` label (the workflow's designed escape hatch) was applied — the check now skips. The alternative was holding the PR until ~2026-06-20; we opted to bypass for a vetted first-party dep.

Local: build ✅ · lint ✅ · format:check ✅ · **423 tests** ✅.

justin-layerv:
Closed the three minor/optional findings from the approving review in `fd638d2`:

1. **baseURL undefined guard** — added `?? ""` to mirror the apiKey guard (no more raw `TypeError` from `.replace` on a stray `undefined`).
2. **Passthrough test coverage** — added adapter-layer assertions for `resolveQURL`, `getQuota`, `mintLink`, and `updateResource` (the methods that previously leaned only on the tool tests).
3. **Dead `tombstone` cast** — left as-is per your call (defensible forward-compat; no consumer reads it), comment already states the SDK error type doesn't expose it.

Local: build ✅ · lint ✅ · format:check ✅ · **427 tests** ✅.

justin-layerv:
Thanks — addressed all four in `af4be7e`:

**1. Node engine mismatch (Medium):** Real mismatch confirmed — `@layervai/qurl`'s `engines.node` is `>=20`. **Bumped this package's `engines.node` to `>=20`** to match (the MCP transitively requires it). Practically a soft constraint for an `npx`-run server — Node 18/19 users would have gotten `EBADENGINE` warnings — but worth declaring honestly. Flagging it raises the minimum supported Node from 18 → 20.

**2. translateError only catches QURLError base (Medium):** Verified — **every SDK error subclass extends `QURLError`**, including transport/timeout:
```
NetworkError → instanceof QURLError: true
TimeoutError → instanceof QURLError: true
ServerError  → instanceof QURLError: true
RateLimitError → instanceof QURLError: true
```
So `translateError` catches network/timeout failures too (status 0). **Added a transport-error test** (`NetworkError` → `QURLAPIError`, statusCode 0, code `network_error`) to lock it for delete-qurl's idempotency.

**3. delete_qurl idempotency on non-JSON 404 (Low):** Verified — the SDK's error parsing falls back to `createError({ status: response.status, code: "unknown", ... })` on a non-JSON error body, so a 404 with an HTML proxy page still yields a `NotFoundError` with **status 404** → `QURLAPIError.statusCode === 404` → the idempotent swallow fires. No regression.

**4. err.code without fallback (Low):** Verified — the SDK types `QURLError.code` as a non-optional `readonly string` and defaults it to `"unknown"` when the API omits it, so `err.code` is always a populated string. The old `?? "unknown"` fallback is redundant.

Local: build ✅ · lint ✅ · format:check ✅ · **428 tests** ✅. CI re-running on `af4be7e`.

justin-layerv:
Addressed the comment nits in \`f0838ec\`:

- **Stale \`request()\` refs:** fixed both — \`index.ts\` deferred-auth note and the \`MISSING_API_KEY_MESSAGE\` doc now point at the lazy \`sdk\` getter / first-call path.
- **\`mintLink\` no-input coverage:** added a one-liner asserting \`mintLink(id)\` passes \`undefined\` through to the SDK.

On the two reasoning items:

- **engines \`>=18\` → \`>=20\` release note:** the bump is forced by the SDK and is documented in the commit body, so release-please's changelog will capture it. I kept it as a non-breaking change rather than \`feat!\` because for an \`npx\`-run MCP server \`EBADENGINE\` is a warning (install still proceeds), not a hard break — but happy to escalate to a \`!\`/BREAKING note in the release if you'd prefer Node 18 users get a louder signal.
- **tombstone (410):** leaving as the noted SDK follow-up per your agreement — zero functional impact today.

Local: build ✅ · lint ✅ · format:check ✅ · **429 tests** ✅.

justin-layerv:
Thanks for the approve. Closed both optional follow-ups in \`87d348c\`:

- **mapResource → handle null:** done. Now drops the \`qurls\` key for both \`undefined\` and a defensive \`null\` (explicit \`=== undefined || === null\` since the repo's \`eqeqeq\` rule forbids \`== null\`). The SDK normalizes null away today, so it's purely defensive.
- **list_qurls \`meta.page_size\`/\`request_id\`:** kept in the output schema rather than trimmed. These are *real API fields* the SDK's \`list()\` simply doesn't surface yet (same shape of gap as \`tombstone\`), so keeping them optional preserves forward-compat — narrowing the MCP's contract would bake the SDK's current limitation in permanently. Filed **layervai/qurl-typescript#133** to surface both the dropped list metadata and the error \`tombstone\` upstream; the adapter picks them up automatically once the SDK exposes them.

Local: build ✅ · lint ✅ · format:check ✅ · **429 tests** ✅.

PR #158, opened 2026-06-13
ci: publish @layervai/qurl-mcp via OIDC trusted publishing

justin-layerv:
## What

Aligns the MCP's npm publish with the `@layervai/qurl` SDK: switches from a `NODE_AUTH_TOKEN` secret to **OIDC trusted publishing**.

The publish job was already 90% there (`environment: npm-publish`, `id-token: write`, `registry-url`, `npm publish --provenance --access public`). This PR:
- **Removes** the `NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}` env from the publish step → npm authenticates via OIDC against the trusted publisher.
- **Pins `npm@11.10.0`** before publish — trusted publishing requires npm ≥ 11.5.1, and Node 22 still bundles npm 10.x. (Same version the SDK uses; old enough to clear the repo's `min-release-age` gate.)

This is the exact setup just **validated** on `@layervai/qurl` (a manual publish there authenticated via OIDC and only failed on "cannot publish over existing version").

## ⚠️ Required before the next release

A **trusted publisher must be configured for `@layervai/qurl-mcp`** on npmjs.com, or the next release's publish will fail auth (the same trap the SDK hit when its trusted publisher was on the wrong package):

- Provider: **GitHub Actions**
- Organization: **layervai** · Repository: **qurl-mcp**
- Workflow filename: **release-please.yml**
- Environment: **npm-publish**

`@layervai/qurl-mcp` already exists on npm (0.4.0), so this is added under the package's **Settings → Trusted Publishers**. Once this lands and a release publishes successfully, the `NPM_TOKEN` repo secret can be removed.

## Note
Only the **npm** publish moves to OIDC. The separate `publish-mcp-registry` job (MCP registry via `mcp-publisher`) already uses OIDC and is unchanged.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -42,15 +42,27 @@ jobs:
         with:
           node-version: "22"
           cache: "npm"
+          # Required for OIDC trusted publishing — npm runs the OIDC exchange
+          # against this registry. setup-node also injects a placeholder
+          # NODE_AUTH_TOKEN, but npm's OIDC auth takes precedence once the
+          # trusted publisher matches (verified on @layervai/qurl).
           registry-url: "https://registry.npmjs.org"
+      # Trusted publishing requires npm >= 11.5.1; Node 22 still bundles npm
+      # 10.x, so pin a known-good CLI (old enough to clear .npmrc's
+      # min-release-age gate) before publishing.
+      - name: Pin npm for trusted publishing
+        run: npm install -g npm@11.10.0
       - name: Install dependencies
         run: npm ci
       - name: Build
         run: npm run build
+      # Trusted publishing (OIDC): no NODE_AUTH_TOKEN. With the job's
+      # `id-token: write`, npm authenticates via the trusted publisher
+      # configured for @layervai/qurl-mcp on npmjs.com (provider GitHub
+      # Actions, repo layervai/qurl-mcp, workflow release-please.yml,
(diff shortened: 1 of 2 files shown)


justin-layerv:
Thanks for the approve. Addressed in \`1b0e856\`, plus fixed the CI failure:

- **`validate` failure (semantic PR title):** the title used scope `(release)`, which isn't in the allowed set (`tools/client/resources/prompts/ci/deps`). Retitled to `ci: …` (no scope) — passes now.
- **Minor #1 — misleading `min-release-age` comment:** you're right, the pin install runs under Node 22's bundled npm **10.x**, which ignores `.npmrc`'s `min-release-age` gate, so age never applies there. Reworded the comment to say that instead of implying 11.10.0 was chosen to clear the gate.
- **Minor #2 — CLAUDE.md doc gap:** added a new **"npm Publishing (Trusted Publishing / OIDC)"** section documenting: no `NODE_AUTH_TOKEN` (and "don't re-add a token when debugging"), the npmjs.com trusted-publisher prerequisite, and the npm≥11.5.1 pin.
- **Minor #3 — pinned npm will age:** noted in both the workflow comment and CLAUDE.md to keep it in sync with the SDK's pinned npm.

**Operational gate** (configuring the trusted publisher for `@layervai/qurl-mcp` before the next release) is now in the PR body *and* CLAUDE.md so it won't get lost.

PR #172, opened 2026-06-27
ci: use shared notify.yml reusable for the Slack build card

justin-layerv:
## What

Converts this repo's inline ~100-line Slack build-notification job into a thin shim over the fleet-shared reusable [`layervai/ops-routines-workflows/.github/workflows/notify.yml@v0.9.0`](https://github.com/layervai/ops-routines-workflows/releases/tag/v0.9.0), mirroring how the `issue-priority` shim was centralized.

- Header preserved: **qURL MCP Build**
- Gates: `{"Lint": "...", "Tests": "...", "Docker": "..."}`
- Run condition (`if:` / `needs:`) unchanged from the inline job.
- The curl/jq Block Kit payload builder and the gate-result aggregation that were duplicated here (3×: `needs:`, per-gate `env:`, aggregation if-chain) now live once in the reusable. Adding a gate is a one-line edit to `needs:` + `gates:`.

## Notes

- Pinned by SHA (`@1536c69 # v0.9.0`), matching the fleet's reusable-ref convention.
- The `notify` job runs only on push (`if: … github.event_name != 'pull_request'`), so it is skipped on this PR — no behavior change to required PR checks.
- `SLACK_WEBHOOK_URL` is passed through as an optional secret; absent ⇒ the reusable no-ops.

Source reusable PR: layervai/ops-routines#80.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -174,110 +174,20 @@ jobs:
             -f description="lint=$LINT_RESULT test=$TEST_RESULT docker=$DOCKER_RESULT" \
             -f target_url="$TARGET_URL"
 
+  # Slack build-completion card. Thin shim over the fleet-shared reusable
+  # in layervai/ops-routines-workflows (public mirror of private
+  # layervai/ops-routines); the curl/jq Block Kit payload builder and the
+  # gate-result aggregation live there now. Adding/removing a gate is a
+  # one-line edit to `needs:` + `gates:` here — no aggregation if-chain.
+  # NOTE: the separate `ci/aggregate` commit-status job above is unrelated.
   notify:
     name: Notify
     needs: [lint, test, docker]
     if: always() && github.event_name != 'pull_request'
-    runs-on: ubuntu-latest
-    timeout-minutes: 5
-    continue-on-error: true
-    steps:
-      - name: Notify Slack
-        env:
-          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
-          TEST_RESULT: ${{ needs.test.result }}
-          LINT_RESULT: ${{ needs.lint.result }}
-          DOCKER_RESULT: ${{ needs.docker.result }}
-          COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
-          GH_SERVER_URL: ${{ github.server_url }}
-          GH_REPOSITORY: ${{ github.repository }}
-          GH_RUN_ID: ${{ github.run_id }}
-          GH_SHA: ${{ github.sha }}
-          GH_REF_NAME: ${{ github.ref_name }}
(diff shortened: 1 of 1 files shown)


justin-layerv:
Thanks for the thorough review. Addressing the recurring verification points (the reusable source isn't visible from the PR diff, so here is the proof against the published `notify.yml@v0.9.0`):

**1. Fail-open on webhook errors — confirmed; the safety net moved intact, not lost.** The reusable's `Notify Slack` step carries *both* layers the inline job had:
- [`continue-on-error: true`](https://github.com/layervai/ops-routines-workflows/blob/v0.9.0/.github/workflows/notify.yml#L76) on the step (a `uses:` caller job can't carry it, so it lives on the step inside the reusable — covers *any* step failure, e.g. `jq` on bad input under `pipefail`), and
- [`curl -sfS … || echo "Slack notification failed (non-critical)"`](https://github.com/layervai/ops-routines-workflows/blob/v0.9.0/.github/workflows/notify.yml#L157) on the request itself.

It also keeps its own [`timeout-minutes: 5`](https://github.com/layervai/ops-routines-workflows/blob/v0.9.0/.github/workflows/notify.yml#L66) (replacing the caller's old cap). So a present-but-failing webhook (non-2xx / DNS / timeout) **cannot** red `main`.

**2. SHA ↔ tag — confirmed.** `gh api repos/layervai/ops-routines-workflows/git/refs/tags/v0.9.0` → `1536c698add5777aa08c781bc674d1eba469b32d`, exactly the pinned SHA.

**3. Card parity — confirmed.** The reusable reconstructs everything from the caller's `github` context: trigger text (Push / PR merge / [`workflow_dispatch`→Manual](https://github.com/layervai/ops-routines-workflows/blob/v0.9.0/.github/workflows/notify.yml#L116)), commit first-line [truncated to 100 chars](https://github.com/layervai/ops-routines-workflows/blob/v0.9.0/.github/workflows/notify.yml#L122) with a [`(no commit message)` fallback](https://github.com/layervai/ops-routines-workflows/blob/v0.9.0/.github/workflows/notify.yml#L123) when `head_commit` is absent, per-gate emoji, and the same Block Kit field layout. The jq payload builder was unit-tested locally against all four gate shapes (success/failure/cancelled/skipped + empty-gates). One *intentional* normalization: the header is now constant across states.

**qurl-mcp specifics (naming nit):** the shim `uses:` the **public mirror** `ops-routines-workflows` (what public repos can call); `layervai/ops-routines` (PR #80) is the **private source of truth** that `sync-public-mirror` copies into that mirror. Both references in the comment/PR body are intentional — calling the mirror, sourced from the private repo. The `ci/aggregate` commit-status job above the notify job is untouched.

